PR #1, opened 2023-11-01
add basic infra definition for a static serving bucket

fredex42:
## What does this change?

Adds basic static serving infrastructure so we can road-test CDN behaviour

## How to test

This has been deployed and hooked fastly onto the bucket as per https://docs.fastly.com/en/guides/amazon-s3


rowannekabalan (approved):
Seen it in action! 👍

PR #2, opened 2023-11-02
Set up datastore and index generation

fredex42:
## What does this change?

- Create a Dynamo table for testing throughput
- Create a script that can be run locally to fill the dynamo table with junk records for testing
- Add library functions to generate index.json from the contents of the dynamo table
- Create a simple lambda function to run the generation on-demand so we can do some basic profiling

## How to test

Run the provided Lambda function via the console and time it.
It should:
- work (i.e. run to completion)
- output timestamps for the various operations it's done
- output data to https://recipes.code.dev-guardianapis.com/index.json.  this can be verified by checking the `last-modified` header in the HTTP response.


fredex42 (commented):
Self-reviewed

PR #4, opened 2023-11-09
Library Function to extract recipe from element

Divs-B:
## What does this change?

A function to fetch the recipe from elements that are present in the block.
Output will contain `recipeUId`, `jsonBlob` and `checksum`(optional)

Reference: https://trello.com/c/ThTd3lIZ/2410-library-function-to-extract-recipe-data-from-an-element-definition

## How to test

At the moment via running the test, `npm test`

Tests should pass


lib/recipes-data/src/lib/extract_recipedata_from_element.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+function parseJsonBlob(canonicalId: string, recipeJson: string | undefined): RecipeOutput {
+  const recipeData = JSON.parse(recipeJson as string) as Record<string, unknown>
+  if (!recipeData.id) {
+    throw new Error("Error! No id present in recipeJson!")
+  } else {
+    return <RecipeOutput>{
+      recipeUId: `${canonicalId}-${recipeData.id as string}`,
+      jsonBlob: JSON.stringify(recipeData)
```
fredex42, line 27: personally I'd just return `recipeData as string` here - it's a bit cheaper as we don't have to put it all back together again from the parsed content

lib/recipes-data/src/lib/extract_recipedata_from_element.ts
```diff
@@ -0,0 +1,30 @@
...
+    return allRecipes.map(recp => parseJsonBlob(webUrl, recp.recipeTypeData?.recipeJson))
+  } else {
+    throw new Error("Error! No recipe is present in elements!")
+  }
+}
+
+function parseJsonBlob(canonicalId: string, recipeJson: string | undefined): RecipeOutput {
+  const recipeData = JSON.parse(recipeJson as string) as Record<string, unknown>
+  if (!recipeData.id) {
+    throw new Error("Error! No id present in recipeJson!")
```
fredex42, line 23: I'd include the `canonicalId` here, so if we see this error in the logs we know where to look

lib/recipes-data/src/lib/extract_recipedata_from_element.ts
```diff
@@ -0,0 +1,30 @@
...
+  jsonBlob: string;
+  checksum?: string;
+}
+
+export function extractRecipeData(webUrl: string, block: Block): RecipeOutput[] {
+  const allRecipes = block.elements.filter(elem => elem.type === ElementType.RECIPE)
+  if (allRecipes.length != 0) {
+    return allRecipes.map(recp => parseJsonBlob(webUrl, recp.recipeTypeData?.recipeJson))
+  } else {
+    throw new Error("Error! No recipe is present in elements!")
```
fredex42, line 16: I don't think that this should be an error condition; I think simply returning an empty array is OK.
(Remember, most of the articles we will be operating on have no recipes in them)

In that case you don't even need the `if`/`else`; you can just `return allRecipes.map(...)` and that will give you an empty output array if the input was empty.

lib/recipes-data/src/res/recipejson-example.json
```diff
@@ -0,0 +1,192 @@
+{
```
fredex42, line 1: Do we still need this test data?

lib/recipes-data/tsconfig.json
```diff
@@ -1,16 +1,25 @@
...
+    "resolveJsonModule": true,
+    "module": "esnext"
+  },
+  "ts-node": {
+    // these options are overrides used only by ts-node
+    // same as the --compilerOptions flag and the TS_NODE_COMPILER_OPTIONS environment variable
+    "compilerOptions": {
+      "module": "commonjs"
+    }
+  }
```
fredex42, line 24: I wouldn't put these here..... they belong in `tsconfig.lib.json` if anywhere - why were they needed?

lib/recipes-data/src/lib/extract_recipedata_from_element.test.ts
```diff
@@ -0,0 +1,375 @@
...
+            mediaApiUri: "https://api.media.gutools.co.uk/images/58c32a98ae4463b5129bf717b1b2312d8ffc0d45",
+            suppliersReference: "Soba noodles with crisp rainbow vegetables and a spicy sesame se",
+            imageType: "Photograph"
+          }
+        }
+      ]
+    }
+    const result = extractRecipeData(webUrl, block)
+    expect(result.length).toEqual(3)
+    expect(result[2].recipeUId).toEqual("http://www.code.dev-theguardian.com/lifeandstyle/2018/jan/05/soba-noodle-salad-vegetables-spicy-sesame-dressing-recipe-thomasina-miers-3") //
```
fredex42, line 199: The ID should not have the full-length URL in it - just the bit from `lifeandstyle/....`.  (mentioned in above comment)

lib/recipes-data/src/lib/extract_recipedata_from_element.test.ts
```diff
@@ -0,0 +1,375 @@
+import {AssetType} from "@guardian/content-api-models/v1/assetType";
+import type {Block} from "@guardian/content-api-models/v1/block";
+import {ElementType} from "@guardian/content-api-models/v1/elementType";
+import {response} from "./../res/recipejson-example.json";
+import {extractRecipeData} from "./extract_recipedata_from_element";
+
+describe("extractRecipeData", () => {
+
+  it("should work when block containing single recipe elements ", () => {
+    const webUrl = response.results[0].webUrl
```
fredex42, line 10: this should be the `id` field not the `webUrl` field.... you might as well just put it in as a static string though and drop the `recipejson-example.json`

fredex42 (commented):
Got a few questions/improvements....

lib/recipes-data/src/lib/extract_recipedata_from_element.ts
Divs-B: `recipeData as string` is not working, it will be nicer to return whole `recipeJson` to `jsonBlob` (discussed with @jonathonherbert) and then to test this by comparing recipeJson and jsonBlob in test file.

lib/recipes-data/src/res/recipejson-example.json
Divs-B: removed 👍

lib/recipes-data/src/lib/extract_recipedata_from_element.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function extractRecipeData(canonicalId: string, block: Block): RecipeOutput[] {
+  const allRecipes = block.elements
+    .filter(elem => elem.type === ElementType.RECIPE)
+    .map(recp => parseJsonBlob(canonicalId, recp.recipeTypeData?.recipeJson))
+  return allRecipes
+}
+
+function parseJsonBlob(canonicalId: string, recipeJson: string | undefined): RecipeOutput {
+  const recipeData = JSON.parse(recipeJson as string) as Record<string, unknown>
```
fredex42, line 19: Sorry to be a pain! I see why you've declared `recipeJson:string|undefined` in the function - to make line 14 easier - but I don't think this is right because on line 18 we are saying that it's OK for `recipeJson` to not be defined then on line 19 we are assuming that it _is_ defined.

I would change line 18 to have `recipeJson:string`, then update lines 13 & 14 to read:

```typescript
   .filter(elem => elem.type===ElementType.RECIPE && !!elem.recipeTypeData?.recipeJson)
   .map(recp => parseJsonBlob(canonicalId, recp.recipeTypeData?.recipeJson as string)
```

The reason that it's OK to do `as string` here is because we have already ensured that it _isn't_ null or undefined in the filter line.  If we're lucky then Typescript will be clever enough to realise this and we can just use `recp.recipeTypeData.recipeJson` without the `?` or forcing the type. But I suspect that it's not clever enough to realise this.

fredex42: The upshot of this change is that if we have an invalid Recipe element - which _shouldn't_ happen, but technically _can_ - then it won't break our lambda, the element will get ignored

fredex42 (changes requested):
just one more thing...

lib/recipes-data/src/lib/extract_recipedata_from_element.ts
Divs-B: Yes this is making more sense, wise way of using the variable with "as string" , thanks @fredex42 for pointing it.

fredex42 (approved):
😁

PR #5, opened 2023-11-10
Fastly purging functions

fredex42:
## What does this change?

Implements https://trello.com/c/u9CjO8o2.

From the card:

**Library function to de-cache a URL from fastly**

Input:

URL to decache (as a string)

Output:

None

You’ll need to look up how to do this in the Fastly docs…

## How to test

There are unit-tests built in, check over the tests and ensure that they pass

Just another brick in the wall

n/a


lib/recipes-data/src/lib/fastly.ts
```diff
@@ -0,0 +1,103 @@
...
+    removeLeadingAndTrailingSlash(contentPath)
+  ].join("/");
+
+  const baseHeaders:Record<string,string> = {
+    "FASTLY-KEY": apiKey,
+    "Accept": "application/json",
+  };
+
+  const headers = purgeType=="hard" ? {
+    "Fastly-Soft-Purge": "1",
```
JustinPinner, line 60: I think this is back-to-front? https://developer.fastly.com/reference/api/purging/#soft-vs-hard-purge

lib/recipes-data/src/lib/fastly.ts
```diff
@@ -0,0 +1,103 @@
...
+ * @param retryCount don't specify this, it's used internally.
+ */
+export async function sendFastlyPurgeRequestWithRetries(contentPath:string, apiKey:string, purgeType?: PurgeType, retryCount?:number):Promise<void>
+{
+  try {
+    return sendFastlyPurgeRequest(contentPath, apiKey, purgeType);
+  } catch(err) {
+    if(err instanceof FastlyError) {
+      const nextRetry = retryCount ? retryCount + 1 : 1;
+      if(nextRetry>MaximumRetries) throw err; //we give up! it ain't gonna work.
```
JustinPinner, line 96: I know `MaximumRetries` has a default in config, what happens if that config is removed? Worth adding a guard here to prevent infinite retries? Maybe it'll collapse with an error anyway if it's `undefined`?

lib/recipes-data/src/lib/fastly.ts
fredex42: ooh, yes! Good catch thanks

lib/recipes-data/src/lib/fastly.ts
fredex42: Good point i'll have a look

PR #6, opened 2023-11-10
S3 writing and deletion functions

fredex42:
## What does this change?

Implements https://trello.com/c/rLxRBgfq

From the card:

**Library function to write the recipe data to the bucket**

Inputs:

output of https://trello.com/c/aHnLbYkp

Outputs:

None

Failures:

should pass on S3 exceptions if they occur. Be prepared to backoff-and-retry in case of 429’s.

Notes:

see https://github.com/guardian/recipes-backend/blob/09802afb0c9d24c03a1331c86e693f6f10750557/lambda/test-indexbuild/src/main.ts#L12 for an example of how to do this

ensure that the ContentType argument is set when writing contents to the bucket (this is picked up and passed on by the CDN)

look up in the AWS docs how to set the cache-control parameters as well. We want to cache for the maximum allowable time, apparently this is 31557600s (1 year)

## How to test

These functions are tested by test-cases in the PR.  Evaluate the tests and ensure that they pass.

Another step on the road

n/a

### DIFF
--- a/lambda/test-indexbuild/src/main.ts
+++ b/lambda/test-indexbuild/src/main.ts
@@ -1,8 +1,8 @@
 import {DynamoDBClient} from "@aws-sdk/client-dynamodb";
 import { PutObjectCommand, S3Client } from "@aws-sdk/client-s3";
 // eslint-disable-next-line import/no-unresolved -- this works from root tsconfig but eslint keeps complaining :-(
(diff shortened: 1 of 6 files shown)


JustinPinner (approved):
LGTM - one minor change to an error message. And assuming FastlyApiKey handling would cause the app to fail earlier if it's missing? That's something for later though :)

PR #7, opened 2023-11-13
Library function to checksum recipe content

fredex42:
## What does this change?

Implements https://trello.com/c/aHnLbYkp

From the card:

**Library function to calculate SHA of the recipe data**
Inputs:
- the output of https://github.com/guardian/recipes-backend/pull/4

Outputs:
- copy of the input data, with the field for the SHA filled in

## How to test

Tests are provided in the PR, evaluate these.

Another step...


rowannekabalan (approved):
Looks good to me 👍 I wonder whether it's better to wait until Divya's PR is merged before merging this? (I worry that we'd otherwise forget to do the #To-Do)

PR #8, opened 2023-11-14
Update removal detection

fredex42:
## What does this change?

- Implements https://trello.com/c/YQmY2nBJ
- Fix eslint config for library tests; loosen the rules on tests files so that they are not a sea of red

From the card:

**Library function to detect taken-down recipes during an update**

Input:
- the output of Library Function to extract all recipes from an article

Output:
- list of articles to take down, in the same format as the input (i.e., same as the input but with articles not to take down removed)

You’ll need to:
- do a DynamoDB Query on the table, for primary key==canonical article ID
- iterate the responses to find the immutable recipe IDs
- compare this with the input list. Any recipe IDs that are present in the input list and not present in the database data should be removed, and therefore make up the response body

## How to test

Automated tests are provided. Look these over and make sure they they are ok!

One more step

### DIFF
--- a/lib/recipes-data/.eslintrc.json
+++ b/lib/recipes-data/.eslintrc.json
@@ -14,6 +14,18 @@
       },
 			"rules": {}
 		},
+    {
+      "files": ["*.test.ts", "*.test.tsx"],
+      "parserOptions": {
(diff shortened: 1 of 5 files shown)


rowannekabalan (approved):
👍

PR #9, opened 2023-11-14
CAPI interface

fredex42:
## What does this change?

Implements https://trello.com/c/JBnAeWBs

## How to test

- Functions are tested with automated tests
- If you deploy it, the function should no longer fail when updates come through but print a single line to the log for each update (takedowns still cause errors)
- Retrievable update acquisition is now implemented

Movin' along


rowannekabalan (approved):
Looks reasonable 👍

PR #10, opened 2023-11-16
Unscrambling types

fredex42:
## What does this change?

Due to the asynchronous development process we've managed to create some overlapping type instances that duplicate functionality.

This PR removes the duplication, creating a single consistent set of types that live in `models.ts`.  References to the fields are also updated.

## How to test

Double-check that the tests make sense and still pass.

More readable code


Divs-B (approved):
Great to have structure aligned with model file. I think makes sense to get checksum work separate (although was optional) from Recipe output

PR #12, opened 2023-11-17
Library function for extracting all recipes.

Divs-B:
## What does this change?

The functionality to extract all recipes present in an article, it will scan all blocks and extract the recipes whether it is present in `main` block or `body` block.

Reference: https://trello.com/c/SzE1z6r8/2413-library-functions-to-extract-all-recipes-from-an-article

## How to test

Via running the test at the moment
`npm test`

The test should pass (present in `extract-recipe.test.ts` and `extract-all-recipes.test.ts`)


lib/recipes-data/src/lib/extract-recipes.ts
```diff
@@ -0,0 +1,38 @@
...
+  const allRecipes = block.elements
+    .filter(elem => elem.type === ElementType.RECIPE)
+    .map(recp => parseJsonBlob(canonicalId, recp.recipeTypeData?.recipeJson as string))
+  return allRecipes
+}
+
+function parseJsonBlob(canonicalId: string, recipeJson: string): RecipeReferenceWithoutChecksum {
+  const recipeData = JSON.parse(recipeJson) as Record<string, unknown>
+  if (!recipeData.id) {
+    throw new Error(`Error! No id present in the recipeJson, canonicalId is ${canonicalId}`)
```
fredex42, line 31: I was chatting with @tonytw1 about whether an exception is the most appropriate way to handle this case and we decided, on balance, that it's probably best to ignore invalid recipes and allow the rest through.

In practise, that means you'll need to change the `parseJsonBlob` function to return `RecipeReferenceWithoutChecksum | null` and return `null` in this case.  Please put a TODO comment in to indicate we should incorporate a metric for failed recipes so we can have an indication of upstream issues.

Once we are returning `RecipeReferenceWithoutChecksum | null` the method caller will get upset.... you'll need to put a `filter` stage in after the `map` that calls this function to filter out all `null` values.  If typescript doesn't realise that `null` is no longer possible then it's safe to cast the value using `as`.

lib/recipes-data/src/lib/extract-recipes.ts
```diff
@@ -0,0 +1,38 @@
...
+  } else {
+    return <RecipeReferenceWithoutChecksum>{
+      recipeUID: `${canonicalId}-${recipeData.id as string}`,
```
fredex42, line 34: This is exactly what was asked for.... however with recent conversations with Basecamp and @tonytw1 the data format changed 😞   Fortunately it makes our lives easier 😁  - we can just set `recipeUID: recipeData.id`

lib/recipes-data/src/lib/utils.ts
```diff
@@ -1,19 +1,32 @@
...
   hasher.update(src.jsonBlob);
   const checksum = hasher.digest("base64url");  //base64 encoding should be more byte-efficient
 
   return {...src, checksum,};
 }
+
+function makeCapiDateTime(from: string): CapiDateTime {
+  const date = new Date(from)
+  return {
+    dateTime: date.getTime().valueOf() as unknown as Int64,// TODO: need to confirm if this is correct approach? In debugging, we can see convresion is happening as expected, example webPublicationDate = { "dateTime": 1515171618000, "iso8601": "2018-01-05T17:00:18Z"}
```
fredex42, line 27: I _think_ it's `new Int64(date.getTime().valueOf())` - remind me to check if it wants the `dateTime` value in milliseconds or nanoseconds.

lib/recipes-data/src/lib/extract-all-recipes.test.ts
```diff
@@ -0,0 +1,201 @@
+import {AssetType} from "@guardian/content-api-models/v1/assetType";
+import type {Content} from "@guardian/content-api-models/v1/content";
+import {ContentType} from "@guardian/content-api-models/v1/contentType";
+import {ElementType} from "@guardian/content-api-models/v1/elementType";
+import {extractAllRecipesFromArticle} from "./extract-recipes";
+import {makeCapiDateTime} from "./utils";
+
+describe("extractAllRecipesFromAnArticle", () => {
+
+  it("should work if main and body both block contains recipes", () => {
```
fredex42, line 10: I'd also put in a checks to see that:
- it's OK if there are no recipes
- multiple body-block recipes passed through correctly
- a single invalid recipe (no id field) is dropped (see later comments about this - currently the code throws an exception)
- a single invalid recipe (invalid json) is dropped

lib/recipes-data/src/lib/utils.ts
Divs-B: as discussed in the chat , implemented the changes and removed as int64 , thx @fredex42

lib/recipes-data/src/lib/extract-recipes.ts
```diff
@@ -0,0 +1,37 @@
...
+import {ElementType} from "@guardian/content-api-models/v1/elementType";
+import type {RecipeReferenceWithoutChecksum} from './models';
+
+export function extractAllRecipesFromArticle(content: Content): Array<RecipeReferenceWithoutChecksum | null> {
+  if (content.type == ContentType.ARTICLE && content.blocks) {
+    const articleBlocks: Blocks = content.blocks
+    const getAllMainBlockRecipesIfPresent: Array<RecipeReferenceWithoutChecksum | null> = extractRecipeData(content.id, articleBlocks.main as Block)
+    const bodyBlocks = articleBlocks.body as Block[]
+    const getAllBodyBlocksRecipesIfPresent: Array<RecipeReferenceWithoutChecksum | null> = bodyBlocks.flatMap(bodyBlock => extractRecipeData(content.id, bodyBlock))
+    return getAllMainBlockRecipesIfPresent.concat(getAllBodyBlocksRecipesIfPresent)
```
fredex42, line 14: would you mind adding on a `.filter(recep=>!!recep)` here and then changing the return type back to `RecipeReferenceWithoutChecksum[]` ? Would make it simpler from calling code

lib/recipes-data/src/lib/extract-recipes.ts
```diff
@@ -0,0 +1,37 @@
...
+  } else {
+    return Array<RecipeReferenceWithoutChecksum>()
+  }
+}
+
+export function extractRecipeData(canonicalId: string, block: Block): Array<RecipeReferenceWithoutChecksum | null> {
+  return block.elements
+    .filter(elem => elem.type === ElementType.RECIPE)
+    .map(recp => parseJsonBlob(canonicalId, recp.recipeTypeData?.recipeJson as string))
+    .filter(recp => !!recp)
```
fredex42, line 24: if this is here, you can set the return type to `RecipeReferenceWithoutChecksum[]` as there is no way it's going to return nulls.

If TS complains about that, you're safe to put `as RecipeReferenceWithoutChecksum[]` onto the end of line 24 to tell it there's no problem

lib/recipes-data/src/lib/extract-recipes.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+function parseJsonBlob(canonicalId: string, recipeJson: string): RecipeReferenceWithoutChecksum | null {
+  const recipeData = JSON.parse(recipeJson) as Record<string, unknown>
+  if (!recipeData.id) {
+    return null //TODO: we should incorporate a metric for failed recipes so we can have an indication of upstream issues.
```
fredex42, line 30: 👍

lib/recipes-data/src/lib/utils.ts
```diff
@@ -1,19 +1,35 @@
...
+export function calculateChecksum(src: RecipeReferenceWithoutChecksum): RecipeReference {
   const hasher = createHash('sha256');
   hasher.update(src.jsonBlob);
   const checksum = hasher.digest("base64url");  //base64 encoding should be more byte-efficient
 
   return {...src, checksum,};
 }
+
+function makeCapiDateTime(from: string): CapiDateTime {
+  const date = new Date(from)
```
fredex42, line 25: if `from` is an ISO datetime string, this should really be `const date = parseISO(from)`

fredex42 (commented):
Great! Just a couple of tiny bits....

lib/recipes-data/src/lib/utils.ts
Divs-B: Good spot. 
changed 👍

lib/recipes-data/src/lib/extract-recipes.ts
Divs-B: Yup, TS complained as you predicted, So had to return with `as` in it..

PR #13, opened 2023-11-20
Bringing together the work into a single lambda function

fredex42:
## What does this change?

Implements https://trello.com/c/msTNa5WY.

With this, the first version of the content pipeline is complete!

### Content update

- extract recipes from article (#12 )
- find recipes to remove/update
- takedown superceded recipes
- upload new versions
- update the index if we've made any content changes

### Content update (retrievable)

- mostly already implemented in #9 
- adding tests, separating from the content-update handler to make the test mocking easier

### Content delete

- already implemented in #11 

## How to test

1. Take a snapshot of the current index state: `curl https://recipes.code.dev-guardianapis.com/index.json | jq > before.json`
2. Go into Composer CODE and create a new article with a recipe in it.  Alternatively, edit the recipe blob of an existing one.
3. Launch your changes
4. Wait for it to propagate.... 60s should be more than enough.  If you're curious, open the Content Platforms space in Central ELK; go to Discover; then open the saved search "Recipe responder"
5. The index should be updated by now.  Take a snapshot of the updated index: `curl https://recipes.code.dev-guardianapis.com/index.json | jq > after.json`
6. Run `diff -u after.json before.json`
7. You should see the timestamp in the index.json has moved and at least one recipe blob has changed.  If you don't, then reload a couple of times more - there are a couple of levels of CDN for it to get through.
8. Retrieve your recipe content from `https://recipes.code.dev-guardianapis.com/content/{checksum-value}`
9. Delete the recipe from your article
10. Re-launch
11. Repeat steps 4-7.  The content should now be gone.
12. Add a recipe again
13. Re-launch
14. Verify it's present in the index
15. Take down the article
16. Repeat steps 4-7.  The content should now be gone. 

Basecamp are able to consume recipes published from Composer

### DIFF
--- a/lambda/recipes-responder/src/takedown_processor.ts
+++ b/lambda/recipes-responder/src/takedown_processor.ts
@@ -1,26 +1,13 @@
 import type {DeletedContent} from "@guardian/content-api-models/crier/event/v1/deletedContent";
 import type {Event} from "@guardian/content-api-models/crier/event/v1/event"
-import {ContentType} from "@guardian/content-api-models/v1/contentType";
-import {awaitableDelay, removeAllRecipesForArticle} from "@recipes-api/lib/recipes-data";
+import {removeAllRecipesForArticle} from "@recipes-api/lib/recipes-data";
 import {DynamoClient} from "./dynamo_conn";
 
-export async function handleTakedown(evt:Event, attempt?:number):Promise<number> {
-  if(evt.payload?.kind=="content") {  //we don't need to handle if it's not an article
-    const content = evt.payload.content;
-    if(content.type==ContentType.ARTICLE) {  //we don't need to handle if it's not an article
-      try {
-        return removeAllRecipesForArticle(DynamoClient, content.id);
-      } catch(err) {
-        if( (attempt ?? 0) > 10) {
-          throw new Error(`Could not remove recipes for taken-down article ${content.id} after 10 attempts, giving up`);
-        }
-        console.error(`WARNING: Could not remove recipes for taken down article ${content.id}: `, err);
-        await awaitableDelay();
-        return handleTakedown(evt, (attempt ?? 0)+1)
-      }
-    }
-  }
-  return 0; //no action => nothing removed
+export async function handleTakedown(evt:Event):Promise<number> {
+  console.log("takedown payload: ", JSON.stringify(evt));
+
+  //there's no payload in the takedown message!
+  return removeAllRecipesForArticle(DynamoClient, evt.payloadId); //evt.payloadId is the canonical article ref that was taken down
 }
 
 // I don't think that these are relevant to us here. So, I'm logging it out to verify that suspicion
--- a/lambda/recipes-responder/src/takedown_processor.ts
+++ b/lambda/recipes-responder/src/takedown_processor.ts
@@ -4,12 +4,12 @@ import {ContentType} from "@guardian/content-api-models/v1/contentType";
 import {awaitableDelay, removeAllRecipesForArticle} from "@recipes-api/lib/recipes-data";
 import {DynamoClient} from "./dynamo_conn";
 
-export async function handleTakedown(evt:Event, attempt?:number):Promise<void> {
+export async function handleTakedown(evt:Event, attempt?:number):Promise<number> {
   if(evt.payload?.kind=="content") {  //we don't need to handle if it's not an article
     const content = evt.payload.content;
     if(content.type==ContentType.ARTICLE) {  //we don't need to handle if it's not an article
       try {
-        await removeAllRecipesForArticle(DynamoClient, content.id);
+        return removeAllRecipesForArticle(DynamoClient, content.id);
       } catch(err) {
         if( (attempt ?? 0) > 10) {
           throw new Error(`Could not remove recipes for taken-down article ${content.id} after 10 attempts, giving up`);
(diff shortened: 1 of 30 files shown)


Divs-B:
In step 1 in the "How to test" instructions, i think it should be `after.json`? 
which is require to do for step 6?

fredex42:
@Divs-B apologies, I had not updated the description. Takedown is working (as is takedown of the final recipe of an article)

lambda/recipes-responder/src/takedown_processor.ts
```diff
@@ -1,29 +1,22 @@
...
+    return 0;
   }
 }
 
-// eslint-disable-next-line @typescript-eslint/require-await -- not implemented yet
-export async function handleDeletedContent(evt:DeletedContent):Promise<void> {
-
-  throw new Error("handleTakedown is not implemented yet")
+// I don't think that these are relevant to us here. So, I'm logging it out to verify that suspicion
+export async function handleDeletedContent(evt:DeletedContent):Promise<number> {
```
JustinPinner, line 19: Deletes are much rarer than takedowns but I believe they can still happen, so this logging may not catch one in the time we give it. In addition to the log action, we could also make a call from here to `handleTakedown` and treat it the same way, at least until we've proved whether deletes still flow out of Composer like they used to.

lambda/recipes-responder/src/update_retrievable_processor.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export async function handleContentUpdateRetrievable(retrievable:RetrievableContent): Promise<number>
+{
+  if(retrievable.contentType!=ContentType.ARTICLE) return 0;  //no point processing live-blogs etc.
+
+  const capiResponse = await retrieveContent(retrievable.capiUrl);
+  switch(capiResponse.action) {
+    case PollingAction.CONTENT_EXISTS:
+      //Great, we have it - but should check if this has now been superceded
+      if(capiResponse.content?.fields?.internalRevision ?? 0 > (retrievable.internalRevision ?? 99) ) {
```
JustinPinner, line 47: The use of `99` here looks suspect to me as it suggests we're assuming there can never be more updates to an item than this, but if it _can_ happen, it almost certainly _will_ at some point.

lib/recipes-data/src/lib/extract-recipes.ts
```diff
@@ -8,30 +9,59 @@ import type {RecipeReferenceWithoutChecksum} from './models';
...
 
-function parseJsonBlob(canonicalId: string, recipeJson: string): RecipeReferenceWithoutChecksum | null {
-  const recipeData = JSON.parse(recipeJson) as Record<string, unknown>
-  if (!recipeData.id) {
-    return null //TODO: we should incorporate a metric for failed recipes so we can have an indication of upstream issues.
+/**
+ * Most recipes have a UUID-style `id` field, so we pass that through.
+ * However some of the ones that were extracted by D&I have a numeric field still.
+ * In that case, we concatenate the canonical ID onto the numeric value and then sha1 the lot.
+ * @param contentIdField
```
JustinPinner, line 34: Might be wise to rename this as it's not the content ID as we typically refer to it - it's the recipe ID it seems

lib/recipes-data/src/lib/s3.ts
```diff
@@ -59,7 +69,7 @@ export async function removeRecipeContent(recipeSHA: string, purgeType?:"soft"|"
   const realAttempt = attempt ?? 1;
 
   const Key = `content/${recipeSHA}`;
-  console.debug(`DEBUG: removeRecipeContent path is s3://${Bucket}/${Key}`);
+  console.debug(`DEBUG: removeRecipeContent path is s3://${Bucket ?? "(no bucket)"}/${Key}`);
```
JustinPinner, line 72: If it's possible that `Bucket` doesn't exist, we should check that and abort this entire feature if it's missing

lib/recipes-data/src/lib/takedown.ts
```diff
@@ -3,12 +3,11 @@ import {recipesforArticle, removeAllRecipeIndexEntriesForArticle, removeRecipe}
 import type { RecipeIndexEntry } from './models';
 import {removeRecipeContent} from "./s3";
 
-async function takeRecipeDown(client: DynamoDBClient, canonicalArticleId: string, recipe: RecipeIndexEntry, removeFromDatabase: boolean):Promise<void>
+async function takeRecipeDown(client: DynamoDBClient, canonicalArticleId: string, recipe: RecipeIndexEntry, unconditional: boolean):Promise<void>
```
JustinPinner, line 6: Can `DynamoDBClient` be relocated to the `./dynamo` import, so there's no need to pass it through these functions?

JustinPinner:
If I'm reading this correctly it feels like we're making a lot of additional work for ourselves in managing "versions" of recipes here. Is this to prevent unwanted overwrites that might result from out-of-sequence updates from Kinesis?

fredex42:
@JustinPinner Exactly.  I realise that out-of-sequence updates are unlikely on LIVE (though possible) but if we ever have to deploy to PREVIEW (again, not going to happen imminently but possible) then they're going to be a fact of life.  I felt it was better to code defensively now rather than have to dig into it again in the future

lambda/recipes-responder/src/takedown_processor.ts
fredex42: The interesting thing here is that the only thing we get is the `DeletedContent` event which contains `aliasPaths` but no canonical paths. I'm therefore not sure _what_ should get deleted in this context (because we are only interested in the canonical path rather than any alias paths).

JustinPinner (approved):
Nice, looking really good I think. Thanks for addressing all my fears! 👍

PR #14, opened 2023-11-23
Infrastructure for serving full REST APIs & curation update endpoint

fredex42:
## What does this change?

- Adds API Gateway going through to a single, backend lambda via proxy integration with mandatory API keys
- POST endpoint to handle incoming curation data

## How to test

When deployed to CODE, take a copy of `curation.json`:
```bash
curl https://recipes.code.dev-guardianapis.com/curation.json > curation.json
```
Verify that it looks like the proper curation document

Now post another json document:
```bash
curl -X POST https://recipes.code.dev-guardianapis.com/api/curation -d@file.json --header 'Content-Type: application/json' --header 'x-api-key: {your-api-key}'
```

You should get a 200 response.

Now read it back:
```bash
curl https://recipes.code.dev-guardianapis.com/curation.json
```

The CDN is set up to reply with stale content during refresh, so you might have to do this a few times in order to get the new data. A cache flush is performed by the lambda, as you can see from the logs.

Curation updates going from minutes/hours (i.e. Ben has to craft the json and send it to us, we have to manually upload) to seconds (Ben sends it directs and tests it)

This is the first POST endpoint on the service. It's protected by an API key which is sufficient for now. We can look into key rotation down the line as the feast app gets more mature


lambda/rest-endpoints/src/app.ts
```diff
@@ -0,0 +1,45 @@
...
+app.set('view engine', 'ejs');
+app.engine('.ejs', ejs);
+
+const router = express.Router();
+router.use(bodyParser.json());
+
+router.post('/api/curation', (req, resp)=>{
+  const textContent = getBodyContentAsJson(req.body);
+
+  if(req.header("Content-Type") != "application/json") {
```
jonathonherbert, line 17: I'll stop mentioning but the non-triple-equalses are making me itchy 😀  – non-blocking, we can sort in one pass with a `format:write` down the line, but enforcing will eliminate a class of bug

lambda/rest-endpoints/src/helpers.ts
```diff
@@ -0,0 +1,12 @@
+export function getBodyContentAsJson(body:unknown): string {
+  if(body instanceof Buffer) {
+    return body.toString('utf-8')
+  } else if(typeof body==='string') {
+    JSON.parse(body);
```
jonathonherbert, line 5: `JSON.parse()` does not have a side effect I know of, is this line necessary?

jonathonherbert (approved):
Tested in CODE, works as expected 🎉 

One line I think is almost certainly unnecessary and some nits, nothing blocking 👍

cdk/lib/recipes-backend.ts
```diff
@@ -61,6 +62,8 @@ export class RecipesBackend extends GuStack {
       default: `/${this.stage}/${this.stack}/recipes-responder/fastly-key`
     })
 
+    const contentUrlBase = this.stage=="CODE" ? "recipes.code.dev-guardianapis.com" : "recipes.guardianapis.com";
```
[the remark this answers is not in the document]
fredex42: meh

cdk/lib/rest-endpoints.ts
```diff
@@ -0,0 +1,68 @@
...
+        STATIC_BUCKET: servingBucket.bucketName,
+        FASTLY_API_KEY: fastlyKey,
+        CONTENT_URL_BASE: contentUrlBase,
+      },
+      fileName: "rest-endpoints.zip",
+      functionName: `recipes-backend-rest-endpoints-${scope.stage}`,
+      handler: "main.handler",
+      memorySize: 128,
+      monitoringConfiguration: {noMonitoring: true},  //for the time being
+      runtime: Runtime.NODEJS_18_X,
```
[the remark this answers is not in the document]
fredex42: I'd rather keep it at 18 for the moment then bring everything to 20 at the same time

lambda/rest-endpoints/src/helpers.ts
fredex42: It will throw an exception if the string is not valid JSON, that's the reason

jonathonherbert: Oh, that's neat. Should probably do it for a Buffer, too, perhaps?

PR #15, opened 2023-11-27
Add CW metric

Divs-B:
## What does this change?

To implement CW metrics to have greater visibility over the work of recipes publisher whether it is working correctly or not.

It should be able to get traces of
1. Succesful recipes
2. Failed recipes
3. Number or rate of index version changes
4. 

Trello card for this: https://trello.com/c/GlPWaSrz/2469-metrics-for-recipes-publisher
co-authored with @fredex42 

## How to test

At present via `npm test`
Need to test it with AWS CW

Test should pass after running test.
Once metrics are live then should be able to get metric on AWS CW as well.

I think, once it is live then to check thoroughly on AWS for CW metrics usages and that if it is not over doing anything and which end up cost-effective?


lib/recipes-data/src/lib/extract-recipes.ts
```diff
@@ -3,31 +3,35 @@ import type {Blocks} from "@guardian/content-api-models/v1/blocks";
...
-    const getAllMainBlockRecipesIfPresent: RecipeReferenceWithoutChecksum[] = extractRecipeData(content.id, articleBlocks.main as Block)
+    const getAllMainBlockRecipesIfPresent = extractRecipeData(content.id, articleBlocks.main as Block)
     const bodyBlocks = articleBlocks.body as Block[]
-    const getAllBodyBlocksRecipesIfPresent: RecipeReferenceWithoutChecksum[] = bodyBlocks.flatMap(bodyBlock => extractRecipeData(content.id, bodyBlock))
-    return getAllMainBlockRecipesIfPresent.concat(getAllBodyBlocksRecipesIfPresent)
+    const getAllBodyBlocksRecipesIfPresent = bodyBlocks.flatMap(bodyBlock => extractRecipeData(content.id, bodyBlock))
+    const recipes = getAllMainBlockRecipesIfPresent.concat(getAllBodyBlocksRecipesIfPresent)
+    const failureCount = recipes.filter(recp => !recp).length
+    await registerMetric("FailedRecipes", failureCount)
+    await registerMetric("SuccessfulRecipes", recipes.length)
```
fredex42, line 18: 👍

fredex42: You should test that `registerMetric()` is properly called in the tests for extract-recipes.ts

package.json
```diff
@@ -11,6 +11,7 @@
   },
   "private": true,
   "dependencies": {
+    "@aws-sdk/client-cloudwatch": "^3.458.0",
     "@aws-sdk/client-dynamodb": "^3.441.0",
     "@aws-sdk/client-s3": "^3.441.0",
```
fredex42, line 16: Let's synchronize the AWS SDK versions to the highest one here (3.458.0)

lambda/recipes-responder/src/main.ts
```diff
@@ -48,11 +49,12 @@ export async function processRecord(r:KinesisStreamRecord):Promise<number> {
...
-export const handler:KinesisStreamHandler = async (event) => {
-  const updatesPerEvent = await Promise.all(event.Records.map((r)=>processRecord(r)));
-  const updatesTotal = updatesPerEvent.reduce((acc, current)=>acc+current, 0);
-  if(updatesTotal>0) {
+export const handler: KinesisStreamHandler = async (event) => {
+  const updatesPerEvent = await Promise.all(event.Records.map((r) => processRecord(r)));
+  const updatesTotal = updatesPerEvent.reduce((acc, current) => acc + current, 0);
+  if (updatesTotal > 0) {
     console.log(`Processed updates for ${updatesTotal} recipes, rebuilding the index json`);
+    await registerMetric("UpdatesTotalOfRecipes", updatesTotal)
```
fredex42, line 57: I believe that there is already a test for `main.ts` - please assert that `registerMetric()` is correctly called in there

fredex42 (commented):
Looks great, just one thing to ensure that `registerMetric` gets called properly. There doesn't appear to be enough logic in the `registerMetric` itself to justify a unit-test

lambda/recipes-responder/src/main.ts
Divs-B: `main.ts` for testing handler is implemented to test `registerMetric` call.

Since direct test for handler (where registerMetric is present) was not available in main.ts hence need to be reviewed whther this implementation is right way to test with right mocking (event, context and callback) for handler method to test KinesisStreamEvent as input. thanks.

Divs-B:
Tested on AWS after deploy on CODE:

image

PR #16, opened 2023-12-05
Add CW alarm

Divs-B:
## What does this change?

Alarm to get notification if lambda runs more than threshold limit means near to its timeout limit.

ref:https://trello.com/c/GlPWaSrz/2469-metrics-for-recipes-publisher

## How to test

Need to deploy on CODE and manually try to trigger alarm by putting some thread.wait may be?
or, by lowering timeout of lambds than its threshold to explicity trigger it?
(need to evaluate on testing scenerios once ready to deploy)

## How can we measure success?

alarm should get triggerd and we should get notification via email

PR #17, opened 2023-12-08
Add CW runtime alarm

Divs-B:
## What does this change?

Alarm to get notification if `recipe-responder` lambda runs more than threshold limit. we taking 75% near to its lambdatimeout as threshold at the moment.

ref:https://trello.com/c/GlPWaSrz/2469-metrics-for-recipes-publisher

(related to PR #16: will be close to avoid any uncertainty due to occurrence of several commits after rebase with `main` )

## How to test

Deploy on CODE

In AWS Content-API account, Cloudwatch -> All alarms -> Search "duration" 

We should be able to see DurationAlarm listed (with word `CODE-recipes-backed`)


fredex42 (approved):
👍 😎

PR #18, opened 2023-12-11
Populate contributors or byline

fredex42:
## What does this change?

We have discussed how to apply "sensible defaults" for contributor lists, both in the pipeline team and with the app developers.

The conclusion is that these fields will be optional coming out of Composer, and if they are null/blank at reception in the recipe backend they will be filled in there. (ref: https://trello.com/c/6wZib7MB)

This PR builds that feature.

- We now (potentially) need to update recipe content.  So, instead of a parse-and-throwaway that we did before it's necessary to hold the whole parsed content while we update and then take a checksum afterwards.  `RecipeReferenceWithoutChecksum` has been updated to accomodate this; when the checksum is calculated, the parsed data is stringified.
- a new library function is created, `updateRecipeFields`, which defines a list of "filter functions" to be applied, in order, to the recipe data.
- filter functions are then provided for both updating the contributors array and byline field.


fredex42:
This is now superceded

PR #22, opened 2024-02-13
Support for freetext and contributor tags from Composer

fredex42:
## What does this change?

Adds support for the Composer changes in https://github.com/guardian/flexible-content/pull/4580.

Specifically, it adds a processing step where we parse the contributors list that comes from Composer as an ADT and put the profile-tag ones into the `contributors` array and the freetext ones into the `byline` array.


lib/recipes-data/src/lib/extract-recipes.ts
```diff
@@ -54,16 +54,51 @@ function determineRecipeUID(recipeIdField: string, canonicalId: string): string
...
+    }
+  });
+
+  return {...parsedRecipe, contributors: contributorTags, byline: freetexts}
+}
+
 function parseJsonBlob(canonicalId: string, recipeJson: string): RecipeReferenceWithoutChecksum | null {
   try {
-    const recipeData = JSON.parse(recipeJson) as Record<string, unknown>
+    const recipeData = JSON.parse(recipeJson) as Record<string, unknown>;
```
emdash-ie, line 91: You would also need to update the type assertion on this line:

```suggestion
    const recipeData = JSON.parse(recipeJson) as (Record<string, unknown> & {contributors: Array<string | Contributor>});
```

PR #29, opened 2024-04-02
Version bucket

jonathonherbert:
## What does this change?

Versions our recipe bucket, to permit the possibility of recovering data if updates to recipes go wrong. May not need it.

## How to test

Ronseal.


jonathonherbert:
Not needed, and in fact not possible given the way we publish recipe artefacts (object name is content hash.)

PR #30, opened 2024-04-04
Re-indexing tool

fredex42:
## What does this change?

- Adds a commandline tool to allow re-indexing of content in the Recipe Backend by grabbing the current published articles from CAPI rather than waiting for them to be re-launched and picked up through Crier in the usual way
- Adds a simple little script to grab the current environment-variable configuration from PROD or CODE environments so that the tool can be easily used
- Remove requirement for node-fetch because we can use built-in `fetch` in Node.js now
- Update tests to mock built-in fetch, also add checks for telemetry calls during take-down that were not previously there
- Add in a telemetry call for a takedown path that was missed

### To Do
- [x] Document

## How to test

- Set up your local environment as per the README
- Choose a recipe ID from CODE. (easiest way is to do `curl https://recipes.code.dev-guardianapis.com/index.json | jq .` and then just pick one
- Download the content `curl https://recipes.code.dev-guaredianapis.com/content/{checksum} | jq . > before.json`
- Run `npm run commandline-reindex --recipe-uid {uid}` and watch the logs
- You should see it load up the article, extract all recipes and re-publish.
- Crucially, the SHA value should not change.
- If the SHA does change, download the new version `curl https://recipes.code.dev-guaredianapis.com/content/{checksum} | jq . > after.json` and then check it at https://www.jsondiff.com/

If you're feeling thorough, then run the test again for a valid CAPI path and a valid Composer article ID, as per the README.

**Finally, don't forget**
- structurise and publish a new recipe on CODE, just to prove that it is still publishing properly

Able to push through changes from downstream of Composer more easily

This is not a tool that should be used often.  As long as it's done in accordance with the instructions, there is minimal risk; notes are in the README for higher-risk operations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 # See http://help.github.com/ignore-files/ for more about ignoring files.
 
+environ-*
+
 # compiled output
 dist
 tmp
--- a/README.md
+++ b/README.md
@@ -5,6 +5,54 @@
 This is a backend service that translates data from the Content API into a format that apps like 
 https://github.com/guardian/ios-feast can use.
 
+# Operations
+
+## Set up for local operations
+
+You need to have a large number of environment variables set for the tools to work. Fortunately there is a script to make setting them up easier.
+1. Get CAPI credentials from Janus.  You'll need the AWS commandline, and `jq` utility, installed; but you should have these already.
+2. Run `STAGE=CODE ./get-local-config.sh` to generate the file `environ-CODE`. Substitute PROD for CODE to get setup for the Production environment (but be careful!)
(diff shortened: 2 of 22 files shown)


jonathonherbert (approved):
Tested locally, works as expected 👍 

I would rather we exercised an endpoint, rather than using local code, to do the ingress and transform! I ran this locally against CODE and wondered why my images were not seeing the change in #31 – of course, I needed to rebase this against `main` to see the difference.

It means you stand a chance of accidentally transforming the entirety of the PROD corpus based on local changes.

Approving to unblock, because we'd really like to transform those images, but yeah flagging that this script flies into the ... ✈️  daaaaanger zone ✈️

PR #27, opened 2024-03-26
Rewrite featured and preview images to fastly URLs on write, if present

jonathonherbert:
## What does this change?

Rewrites image URLs for featured and preview images to URLs that use the [fastly resizer service](https://github.com/guardian/fastly-image-service/tree/34399e065bc85b8ca4d5adeeca02b4b6404eaa98), if it's possible (the image must be present and have a media and crop ID, or a URL from which it's possible to extract a crop ID.)

in the future, we'd like to add a `templateUrl` property to allow the app to template in its own parameters, but in the short term this should get us off the ground.

Only newly edited recipes will see this change.

## How to test

- The automated tests should pass. I've thought of a few unhappy paths – perhaps there are more. See the test descriptions for details.
- Deploy to CODE, and make a change to a recipe. You should see the appropriate fastly URLs for the recipe images in backend. They should work!

- [x] Tested in CODE with 1 ([composer](https://composer.code.dev-gutools.co.uk/content/65e5ce818f086762e6507c74), [recipe](https://recipes.code.dev-guardianapis.com/content/chTD37k2MCG8cXSSqKVbV3vpdwBXYCa9ec7hD-hGT6E)), 2 ([composer](https://composer.code.dev-gutools.co.uk/content/5a16b8e793c7ae06fe472272), [recipe](https://recipes.code.dev-guardianapis.com/content/iz4V62YerXHH3GgVx05lTpw_ffEVBPQERUTYvbM9Vko)) – note the differing image URLs. The recipe-backend images should all be i.guim.co.uk.

## How can we measure success?

Smaller images for the recipes app.

PR #31, opened 2024-04-09
Bump image width

jonathonherbert:
## What does this change?

After #27, bumps the image width to 1600px for both image types, to better fit featured recipes on larger devices. 

In conversation with @robindorpe-guardian, we think this should be a manageable size for the app.

### DIFF
--- a/lib/recipes-data/src/lib/config.ts
+++ b/lib/recipes-data/src/lib/config.ts
@@ -29,6 +29,6 @@ export const TelemetryXAR = process.env["TELEMETRY_XAR"]
 export const TelemetryTopic = process.env["TELEMETRY_TOPIC"]
 
 //Used by content transforms
-export const PreviewImageWidth = process.env["PREVIEW_IMAGE_WIDTH"] ? parseInt(process.env["PREVIEW_IMAGE_WIDTH"]) : 300;
-export const FeaturedImageWidth = process.env["FEATURED_IMAGE_WIDTH"] ? parseInt(process.env["FEATURED_IMAGE_WIDTH"]) : 700;
+export const PreviewImageWidth = process.env["PREVIEW_IMAGE_WIDTH"] ? parseInt(process.env["PREVIEW_IMAGE_WIDTH"]) : 1600;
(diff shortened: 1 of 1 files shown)


emdash-ie (approved):
Excellent, happy to be sending bigger images that we’re happy with! The env vars aren’t used anywhere that needs updating, are they?

lib/recipes-data/src/lib/config.ts
```diff
@@ -29,6 +29,6 @@ export const TelemetryXAR = process.env["TELEMETRY_XAR"]
 export const TelemetryTopic = process.env["TELEMETRY_TOPIC"]
 
 //Used by content transforms
-export const PreviewImageWidth = process.env["PREVIEW_IMAGE_WIDTH"] ? parseInt(process.env["PREVIEW_IMAGE_WIDTH"]) : 300;
-export const FeaturedImageWidth = process.env["FEATURED_IMAGE_WIDTH"] ? parseInt(process.env["FEATURED_IMAGE_WIDTH"]) : 700;
+export const PreviewImageWidth = process.env["PREVIEW_IMAGE_WIDTH"] ? parseInt(process.env["PREVIEW_IMAGE_WIDTH"]) : 1600;
+export const FeaturedImageWidth = process.env["FEATURED_IMAGE_WIDTH"] ? parseInt(process.env["FEATURED_IMAGE_WIDTH"]) : 1600;
```
JustinPinner, line 33: Are there any lower-bounds for what could be set in the env vars? E.g. if we have an env var that says 300 we'd still select that over 1600?

jonathonherbert:
_> Excellent, happy to be sending bigger images that we’re happy with! The env vars aren’t used anywhere that needs updating, are they?

Not yet – that was the intent of putting them here, though! Happy to follow up with a PR that threads through some stack params, e.g.

lib/recipes-data/src/lib/config.ts
jonathonherbert: I'm tempted to say caveat emptor on the env vars, which would arrive from the stack – but we could add this if we thought it was a sensible safeguard.

PR #32, opened 2024-04-11
Use correct dimensions for master crop image

jonathonherbert:
## What does this change?

This PR corrects some faulty logic that means a few recipe image urls are giving 404s.

## Why?

At the moment, in our image resizer URLs transform (see #27), we're using the `/master` object path, which only contains the full-size image – but we're not using the master crop width to determine the URL. For example, at the moment we generate https://i.guim.co.uk/img/media/7f96c515c4e320b8ded848f23ffdef8bd311fcad/245_1381_2750_2751/master/2000.jpg?width=1600&dpr=1&s=none, because the `width` property of the image object is `2000`, but we should generate https://i.guim.co.uk/img/media/7f96c515c4e320b8ded848f23ffdef8bd311fcad/245_1381_2750_2751/master/2750.jpg?width=1600&dpr=1&s=none.

This mostly works, because VCL has a fallback when it can't find the master – it will try progressively smaller crop sizes, based on the aspect ratio it derives from the crop ID (format `x_y_width_height`, e.g. `245_1381_2750_2751` above) – [VCL here.](https://github.com/guardian/fastly-image-service/blob/34399e065bc85b8ca4d5adeeca02b4b6404eaa98/fastly-io_guim_co_uk/src/main/resources/varnish/main.vcl#L161) However, in the image URL above, the image is not quite square.

This means the fastly image resizer assumes an aspect ratio that is not _quite_ 1:1, and looks for an image in the crops bucket called `1999.jpg`, which doesn't exist.

To fix, we get the master width from the crop id in the URL, and use that, together with the `/master` object path, to look for the right master image every time.

This PR also avoids exercising failover logic when we are missing a `width` or `cropId` property in the image object – we do not use those properties any more, instead deriving everything from the image URL.

## How to test

- The automated tests should pass. I've adjusted the original transform tests to account for the new behaviour, and added new data to cover extracting the relevant params from the image url.
- Run a reindex (#30) in CODE. Image URLs should work as expected. Running the same code in PROD should correct a few recipes with known problems – e.g. at least one of the images in [1](https://recipes.guardianapis.com/api/content/by-uid/961c127e986b4a0e95aa358d3e62cf63), [2](https://recipes.guardianapis.com/api/content/by-uid/d577a532513aa0a6030b4e72dc03eb5b326b586d), [3](https://recipes.guardianapis.com/api/content/by-uid/838ccd441ce422c71a6169453180cea9aa0b6e97) are currently broken. (You'll need an API key for recipes-backend for these links, happy to pair to create one.)
- We can test images with https://github.com/guardian/recipe-tools/pull/5. The above pieces are now fixed!

```
Recipe 961c127e986b4a0e95aa358d3e62cf63
Featured image https://i.guim.co.uk/img/media/7f96c515c4e320b8ded848f23ffdef8bd311fcad/0_470_3761_4702/master/3761.jpg?width=1600&dpr=1&s=none: 200
Preview image https://i.guim.co.uk/img/media/7f96c515c4e320b8ded848f23ffdef8bd311fcad/245_1381_2750_2751/master/2750.jpg?width=1600&dpr=1&s=none: 200
Recipe d577a532513aa0a6030b4e72dc03eb5b326b586d
Featured image https://i.guim.co.uk/img/media/a51327900002017a5f09af4b0c711d164a655e4f/0_888_3731_3731/master/3731.jpg?width=1600&dpr=1&s=none: 200
Preview image https://i.guim.co.uk/img/media/a51327900002017a5f09af4b0c711d164a655e4f/0_888_3731_3731/master/3731.jpg?width=1600&dpr=1&s=none: 200
Recipe 838ccd441ce422c71a6169453180cea9aa0b6e97
Featured image https://i.guim.co.uk/img/media/652ac636ff7f245f37d135ccc68ead315e2a289b/0_854_3731_3731/master/3731.jpg?width=1600&dpr=1&s=none: 200
Preview image https://i.guim.co.uk/img/media/652ac636ff7f245f37d135ccc68ead315e2a289b/0_854_3731_3731/master/3731.jpg?width=1600&dpr=1&s=none: 200
```

### DIFF
--- a/lib/recipes-data/src/lib/utils.ts
+++ b/lib/recipes-data/src/lib/utils.ts
@@ -41,7 +41,7 @@ export const extractCropDataFromGuimUrl = (
 	url: string,
 ): { mediaId: string; cropId: string; width: number } | undefined => {
 	const match = url.match(
-		/https:\/\/.*?\/(?<mediaId>.*?)\/(?<cropId>\d{1,4}_\d{1,4}_(?<width>\d{1,4})_\d{1,4})\/(?<fileName>.*?)$/,
+		/https:\/\/.*\/(?<mediaId>.*?)\/(?<cropId>\d{1,4}_\d{1,4}_(?<width>\d{1,4})_\d{1,4})\/(?<fileName>.*?)$/,
 	);
 
 	if (!match?.groups) {
--- a/lib/recipes-data/src/lib/utils.ts
+++ b/lib/recipes-data/src/lib/utils.ts
@@ -39,22 +39,23 @@ export function nowTime():Date {
 
 export const extractCropDataFromGuimUrl = (
 	url: string,
-): { cropId: string; width: number } | undefined => {
+): { mediaId: string; cropId: string; width: number } | undefined => {
 	const match = url.match(
-		/https:\/\/.*?\/(?<id>.*?)\/(?<cropId>\d{1,4}_\d{1,4}_(?<width>\d{1,4})_\d{1,4})\/(?<fileName>.*?)$/,
+		/https:\/\/.*?\/(?<mediaId>.*?)\/(?<cropId>\d{1,4}_\d{1,4}_(?<width>\d{1,4})_\d{1,4})\/(?<fileName>.*?)$/,
 	);
 
 	if (!match?.groups) {
 		return;
 	}
 
-	const { cropId, width } = match.groups;
+	const { mediaId, cropId, width } = match.groups;
 
-	if (!cropId || !width) {
+	if (!mediaId || !cropId || !width) {
 		return;
 	}
 
 	return {
+    mediaId,
 		cropId,
 		width: parseInt(width),
 	};
--- a/lib/recipes-data/src/lib/utils.ts
+++ b/lib/recipes-data/src/lib/utils.ts
@@ -37,7 +37,25 @@ export function nowTime():Date {
   return new Date();
 }
 
-export const extractCropIdFromGuimUrl = (url: string): string | undefined =>
-	url.match(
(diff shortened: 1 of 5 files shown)


lib/recipes-data/src/lib/utils.ts
```diff
@@ -37,7 +37,26 @@ export function nowTime():Date {
...
 
-export const extractCropIdFromGuimUrl = (url: string): string | undefined =>
-	url.match(
-		/https:\/\/.*?\/(?<id>.*?)\/(?<cropId>\d{1,4}_\d{1,4}_\d{1,4}_\d{1,4})\/(?<fileName>.*?)$/,
-	)?.groups?.cropId;
+export const extractCropDataFromGuimUrl = (
+	url: string,
+): { mediaId: string; cropId: string; width: number } | undefined => {
+	const match = url.match(
+		/https:\/\/.*?\/(?<mediaId>.*?)\/(?<cropId>\d{1,4}_\d{1,4}_(?<width>\d{1,4})_\d{1,4})\/(?<fileName>.*?)$/,
```
JustinPinner, line 44: I may be wrong (_reading_ regexes is hard!) but is the `<fileName>` group needed? Not a blocker, just thought I'd ask :)

JustinPinner (approved):
Tests look good, rationale seems sound. Approving subject to other opinions 😄 👍

lib/recipes-data/src/lib/utils.ts
jonathonherbert: It's not! I included it to illustrate the shape of the URL, if that makes sense. If it's unnecessary or distracting, we can remove.

PR #33, opened 2024-04-12
Respect image extension when generating fastly URLs

jonathonherbert:
## What does this change?

Respects the image extension when generating Fastly URLs.

Only spotted by scraping all our image links with https://github.com/guardian/recipe-tools/pull/5 and finding the one image in ~2000 that was a `.png` 🤦 

## How to test

- The automated tests should pass. I've added two to cover this case.
- Once merged, reindex recipe `6f8d1bcb6fab4d98ab138e35ca542478` in PROD. Both its featured and preview image should no longer 404.


emdash-ie (approved):
Looks good!

PR #34, opened 2024-04-22
Support timed "launch" of curation data

fredex42:
## What does this change?

- Updates the recipe backend API to support dated curation delivery.
- Adds a lambda function to "launch" todays curation by copying it over the default path
  - Also detects when an update is made to todays curation and "re-launches" it

### New endpoints/arguments:

**POST** https://recipes.code.dev-guardianapis.com/api/curation/{region}/{variant}?date={YYYY-MM-DD} [authenticated]
**GET** https://recipes.code.dev-guardianapis/{region}/{variant}/{YYYY-MM-DD}/curation.json [open]

## How to test

1. POST some JSON to a 'test' region/variant combination, e.g. `region-andy-test`/`variant-one` using TOMORROW'S date.  Use a name that nobody else will use as you'll need to go back tomorrow.
2. Test that you can access it with the GET request above
3. Test that **GET** https://recipes.code.dev-guardianapis/{region}/{variant}/curation.json returns you 403
4. **POST** some _different_ JSON to the same region/variant combination, using TODAY'S date
5. **GET** https://recipes.code.dev-guardianapis/{region}/{variant}/curation.json should now return you this JSON (you might need to try a couple of times to allow for CDN lag)
6. Test the same GET tomorrow.  It should now return you the JSON you originally uploaded with tomorrow's date.

Able to support future-launching of fronts

There is currently no protection against one user over-writing another's changes.  This will come with the Fronts Tool work.

### DIFF
--- a/cdk/lib/recipes-backend.ts
+++ b/cdk/lib/recipes-backend.ts
@@ -174,6 +174,11 @@ export class RecipesBackend extends GuStack {
           effect: Effect.ALLOW,
           actions: ["s3:PutObject","s3:GetObject"],
           resources: [serving.staticBucket.bucketArn + "/*"]
+        }),
+        new PolicyStatement({
+          effect: Effect.ALLOW,
+          actions: ["s3:ListBucket"],
+          resources: [serving.staticBucket.bucketArn]
         })
       ],
       memorySize: 128,
--- a/cdk/lib/recipes-backend.ts
+++ b/cdk/lib/recipes-backend.ts
@@ -181,7 +181,7 @@ export class RecipesBackend extends GuStack {
           resources: [serving.staticBucket.bucketArn]
         })
       ],
-      memorySize: 128,
+      memorySize: 256,
       monitoringConfiguration: {
         noMonitoring: true,
       },
--- a/cdk/lib/recipes-backend.ts
+++ b/cdk/lib/recipes-backend.ts
@@ -189,6 +189,7 @@ export class RecipesBackend extends GuStack {
       environment: {
         STATIC_BUCKET: serving.staticBucket.bucketName,
         FASTLY_API_KEY: fastlyKeyParam.valueAsString,
+        CONTENT_URL_BASE: contentUrlBase,
       }
     });
 
--- a/cdk/lib/recipes-backend.ts
+++ b/cdk/lib/recipes-backend.ts
@@ -159,7 +159,7 @@ export class RecipesBackend extends GuStack {
     durationAlarm.addAlarmAction(new SnsAction(nonUrgentAlarmTopic));
 
     const publishTodaysCurationLambda = new GuScheduledLambda(this, "PublishTodaysCuration", {
-      app: "recipies-publish-todays-curation",
+      app: "recipes-publish-todays-curation",
       architecture: Architecture.ARM_64,
       fileName: "publish-todays-curation.zip",
       functionName: `PublishTodaysCuration-${props.stage}`,
--- a/cdk/lib/recipes-backend.ts
+++ b/cdk/lib/recipes-backend.ts
@@ -1,3 +1,4 @@
+import {GuScheduledLambda} from "@guardian/cdk";
 import type {GuStackProps} from "@guardian/cdk/lib/constructs/core";
 import {GuParameter, GuStack} from "@guardian/cdk/lib/constructs/core";
 import {GuLambdaFunction} from "@guardian/cdk/lib/constructs/lambda";
@@ -6,8 +7,10 @@ import {StreamRetry} from "@guardian/cdk/lib/utils/lambda";
 import {type App, aws_sns, Duration} from "aws-cdk-lib";
 import {Alarm, ComparisonOperator, TreatMissingData, Unit} from "aws-cdk-lib/aws-cloudwatch";
 import {SnsAction} from "aws-cdk-lib/aws-cloudwatch-actions";
+import {Schedule} from "aws-cdk-lib/aws-events";
 import {Effect, PolicyStatement} from "aws-cdk-lib/aws-iam";
 import {Architecture, Runtime} from "aws-cdk-lib/aws-lambda";
+import {LambdaDestination} from "aws-cdk-lib/aws-s3-notifications";
 import {DataStore} from "./datastore";
 import {ExternalParameters} from "./external_parameters";
 import {RestEndpoints} from "./rest-endpoints";
@@ -47,6 +50,9 @@ export class RecipesBackend extends GuStack {
         })
       ]
     });
+    const externalParameters = new ExternalParameters(this, "externals");
+    //const urgentAlarmTopic = aws_sns.Topic.fromTopicArn(this, "urgent-alarm", externalParameters.urgentAlarmTopicArn.stringValue);
+    const nonUrgentAlarmTopic = aws_sns.Topic.fromTopicArn(this, "nonurgent-alarm", externalParameters.nonUrgentAlarmTopicArn.stringValue);
 
     //This is a nicer way to pick up the stream name - but CDK won't compile
     //when using the name token for the kinesis stream name below.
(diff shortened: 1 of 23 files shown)


cdk/lib/recipes-backend.ts
```diff
@@ -47,6 +50,9 @@ export class RecipesBackend extends GuStack {
         })
       ]
     });
+    const externalParameters = new ExternalParameters(this, "externals");
+    //const urgentAlarmTopic = aws_sns.Topic.fromTopicArn(this, "urgent-alarm", externalParameters.urgentAlarmTopicArn.stringValue);
```
jonathonherbert, line 54: ✂️

cdk/lib/recipes-backend.ts
```diff
@@ -147,13 +153,55 @@ export class RecipesBackend extends GuStack {
...
+          resources: [serving.staticBucket.bucketArn]
+        })
+      ],
+      memorySize: 256,
+      monitoringConfiguration: {
+        noMonitoring: true,
+      },
+      rules: [{
+        schedule: Schedule.cron({hour: "0", minute: "1"}),
+        description: "Update Feast app daily curation at midnight"
```
jonathonherbert, line 190: A quick note that we may need different schedules for different editions. I wonder if, in future, we might delegate to clients, which will have their own opinion about when midnight is. This'll be fine for now.

cdk/lib/rest-endpoints.ts
```diff
@@ -47,7 +47,7 @@ export class RestEndpoints extends Construct {
       fileName: "rest-endpoints.zip",
       functionName: `recipes-backend-rest-endpoints-${scope.stage}`,
       handler: "main.handler",
-      memorySize: 128,
+      memorySize: 256,
```
jonathonherbert, line 50: Interesting. Any metrics or observations suggesting this was struggling with less memory?

lambda/publish-todays-curation/src/curation.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    expect(response).toBeNull();
+
+    expect(s3Mock.commandCalls(HeadObjectCommand).length).toEqual(1);
+    const c = s3Mock.commandCalls(HeadObjectCommand)[0];
+    const arg = c.firstArg as HeadObjectCommand;
+    expect(arg.input.Bucket).toEqual("TestBucket");
+    expect(arg.input.Key).toEqual("some-region/some-variant/2024-03-03/curation.json");
+  });
+
+  it("curation.should pass on any other error as an exception", async () => {
```
jonathonherbert, line 126:
```suggestion
  it("should pass on any other error as an exception", async () => {
```

lambda/publish-todays-curation/src/curation.test.ts
```diff
@@ -0,0 +1,214 @@
...
+      .resolvesOnce({})
+      .rejects(new NotFound({$metadata: {}, message: ""}));
+
+    const result = await validateAllCuration(Today, false);
+
+    expect(s3Mock.commandCalls(HeadObjectCommand).length).toEqual(4);
+    for(let i=0; i<4; i++) {
+      const req = s3Mock.commandCalls(HeadObjectCommand)[i].firstArg.input as HeadObjectCommandInput;
+      expect(req.Bucket).toEqual(Bucket);
+      switch(i) {
```
jonathonherbert, line 156: definitely non-blocking: perhaps better expressed by iterating over an array of things we'd like to assert?

lambda/publish-todays-curation/src/curation.ts
```diff
@@ -0,0 +1,132 @@
...
+  month: number;
+  day: number;
+}
+
+const KnownRegions = [
+  "northern",
+  "southern"
+];
+
+const KnownVariants = [
```
jonathonherbert, line 21: Shall we keep this as Front, or does Variant mean something else in this case? I'd recommend having things line up just to keep things as straightforward as poss down the pipeline, but there may be good reasons

lambda/publish-todays-curation/src/curation.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export interface CurationPath {
+  region: string;
+  variant: string;
+  year: number;
+  month: number;
+  day: number;
+}
+
+const KnownRegions = [
```
jonathonherbert, line 16: As w/ Fronts, this could be 'edition' rather than region, again there may be reasons and it does seem like these map straightforwardly to regions – for now 🌏

lambda/publish-todays-curation/src/curation.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  const allCurations = await Promise.all(promises);
+  return allCurations.filter((c)=>!!c) as CurationPath[];
+}
+
+export function generatePath(region:string, variant:string, date:Date) {
+  return `${region}/${variant}/${format(date, DateFormat)}/curation.json`;
+}
+
+function zeroPad(num:number, places:number) {
```
jonathonherbert, line 48: reminds me of [this,](https://qz.com/646467/how-one-programmer-broke-the-internet-by-deleting-a-tiny-piece-of-code) he he he

lambda/rest-endpoints/src/helpers.ts
```diff
@@ -10,3 +10,22 @@ export function getBodyContentAsJson(body:unknown): string {
     throw new Error("Did not recognise the body content as json or json-like");
   }
 }
+
+export function validateDateParam(dateParam:string):Date|null {
```
jonathonherbert, line 14: Could we not use e.g. `new Date("YYYY-MM-DD")` and then assert it's not `typeof InvalidDate`? The `Date` constructor uses [`Date.parse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse), which [uses a date time string format that's compatible IIUC](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format)

lambda/rest-endpoints/src/submit-data.test.ts
```diff
@@ -0,0 +1,68 @@
+import {PutObjectCommand, S3Client} from "@aws-sdk/client-s3";
+import {mockClient} from "aws-sdk-client-mock";
+import {importNewData} from "./submit-data";
+import {sendFastlyPurgeRequestWithRetries} from "@recipes-api/lib/recipes-data";
+import MockedFn = jest.MockedFn;
+import Mock = jest.Mock;
+
+const s3Mock = mockClient(S3Client);
+//const dynamoMock = mockClient(DynamoDBClient);
```
jonathonherbert, line 9: ✂️

jonathonherbert (commented):
This looks great, a few comments! Testing in CODE now before ✅

cdk/lib/recipes-backend.ts
fredex42: Yeah I'd like it with clients..... but as discussed this is OK for now.

cdk/lib/rest-endpoints.ts
fredex42: Yeah the logs show the total amount of used RAM, and I saw it was up fairly close to 128. Also, bumping RAM gives a higher CPU cap => faster execution

lambda/publish-todays-curation/src/curation.test.ts
fredex42: probably yes 😁

lambda/publish-todays-curation/src/curation.ts
fredex42: 😆 that's why i write it myself!

lambda/rest-endpoints/src/helpers.ts
fredex42: Interesting question.... we discussed ease of use vs strict validation vs library/core and decided that there wasn't much in it. Interestingly the Date.parse logic seems to validate in the same way as this code.... i.e. 31st Feb _is_ treated as a valid date... the 2nd (or 3rd) March, depending on leap-year

jonathonherbert (approved):

PR #36, opened 2024-05-03
Connect to facia

fredex42:
## What does this change?

Sets up recipes-backend to be able to receive content sent from Fronts Tool (aka Facia) as per https://github.com/guardian/facia-tool/pull/1565

## How to test

- Deploy to CODE
- Ensure that the SQS queue created by CDK here is subscribed to the Feast output topic in the CMS Fronts account
- Ensure that the latest version of https://github.com/guardian/facia-tool/pull/1565 is running on facia
- Publish Feast curation from facia CODE
- Watch the new lambda receive it
- It should be available at the usual endpoints

Able to publish Feast curation to CODE from Facia

- At present, this will over-write curation data already in place. Need to look at protecting mise-en-place content from overwriting.


JustinPinner:
> Need to look at protecting mise-en-place content from overwriting.

Is this still a concern? If so should we mark this PR as draft, or apply the do-not-merge tag to it? I haven't started the review yet so this question may answer itself in due course :)

lib/facia/src/lib/facia-models.ts
```diff
@@ -0,0 +1,77 @@
+import { z } from "zod";
+
+//This should mirror FeastAppModel in Facia backend
+export const RecipeIdentifier = z.object({
+  id: z.string()
+})
+
+export type RecipeIdentifier = z.infer<typeof RecipeIdentifier>;
```
JustinPinner, line 8: I'm not familiar with what `zod` does or how, but will it be confusing to others in future if there are multiple exports with the same name?

JustinPinner: I guess what I _meant_ to ask was is it necessary to export both the `const` and the `type` definitions? Would we be importing `type`s primarily? Will read up in case I can answer this for myself :)

jonathonherbert (commented):
Looks great! A few comments while I wrap up the SNS connection infra for testing end-to-end.

jonathonherbert (approved):
Approving to unblock, @fredex42 might be worth taking a look at the changes I've made in testing!

A few non-blocking comments mainly about structure, I think we can just crack on 😀

lib/facia/src/lib/facia-models.ts
fredex42: `zod` is a library that does runtime JSON validation for Typescript - it's been adopted in CoPro to ensure that the json objects really do match the type signatures so we are using it here as well!

This is actually the recommended way of using it, despite it looking a bit weird.  The reason you have both a `const` and a `type` with the same name is so that you can do e.g. `RecipeIdentifier.parse(untypedJsonObject)` and get a `RecipeIdentifier` instance from it. So it kinda looks like the Scala companion-object pattern from the POV of the caller

PR #38, opened 2024-06-06
Add templateUrl field so frontend can get needed image size

michaelclapham:
## What does this change?
Adds a new field with a placeholder of #{width} so that the frontend can request the recipe image at the desired width needed based on the context.

## How to test

- [x] The automated tests should pass.
- [ ] Test in CODE with a consuming device. Images should appear as expected.

Data usage for app users goes down by a factor of 10.

### DIFF
--- a/lib/recipes-data/src/lib/models.ts
+++ b/lib/recipes-data/src/lib/models.ts
@@ -121,6 +121,7 @@ export function recipeReferenceFromIndexEntry(entry: RecipeIndexEntry, jsonBlob:
 
 export type RecipeImage = {
 	url: string;
+  templateUrl?: string; // Contains #{width} so that device can request image at needed size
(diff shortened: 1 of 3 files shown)


jonathonherbert:
Tests added, so this is good to 🚢  – before we do that, @michaelclapham would you like to test this in a CODE environment with a running application? Happy to help!

michaelclapham:
> Tests added, so this is good to 🚢 – before we do that, @michaelclapham would you like to test this in a CODE environment with a running application? Happy to help!

Thanks for adding the tests!! Yes please I could do with some help testing this 🙏

jonathonherbert (approved):

jonathonherbert:
We've reindexed in CODE and can see this present on recipes, so happy to merge 👍

PR #43, opened 2024-06-24
Minor template URL changes after feedback from iOS

jonathonherbert:
## What does this change?

A few minor changes re: feedback –
- `templateUrl` -> `templateURL`, as this is what the (in production) iOS app already uses
- added a `quality` template param

Also a minor refactor as the arity of some functions was quite large, which merited a conversion to an options object.

## How to test

- [x] The automated tests should pass.
- [ ] Deploy to CODE, reindex content, and verify that things are as expected with the Feast team.

### DIFF
--- a/lib/recipes-data/src/lib/transform.ts
+++ b/lib/recipes-data/src/lib/transform.ts
@@ -2,30 +2,50 @@ import { FeaturedImageWidth, ImageDpr, PreviewImageWidth } from './config';
 import type { Contributor, RecipeImage } from './models';
 import { extractCropDataFromGuimUrl } from './utils';
 
-const getFastlyUrl = (
-	imageId: string,
-	cropId: string,
-	dpr: number,
-	desiredWidth: number,
-	originalWidth: number,
-  extension: string
-) => `https://i.guim.co.uk/img/media/${imageId}/${cropId}/master/${originalWidth}.${extension}?width=${desiredWidth}&dpr=${dpr}&s=none`;
+const getFastlyUrl = ({
+	imageId,
+	cropId,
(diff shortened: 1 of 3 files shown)


lib/recipes-data/src/lib/transform.ts
```diff
@@ -2,30 +2,50 @@ import { FeaturedImageWidth, ImageDpr, PreviewImageWidth } from './config';
...
+	originalWidth,
+	extension,
+}: {
+	imageId: string;
+	cropId: string;
+	dpr: number;
+	desiredWidth: number;
+	originalWidth: number;
+	extension: string;
+}) =>
```
emdash-ie, line 19: It’s a shame this syntax is so verbose: it’s definitely nice to use an object for the calling code, but duplicating all the variable names here is a little annoying.

emdash-ie (approved):
Looks good! What’s the reason for changing the casing of `templateURI`? (I prefer the new casing, but just wondering if it’s for compatibility with something else)

lib/recipes-data/src/lib/transform.ts
jonathonherbert: Yes, it's a bit awkward! It's the only way to write something close to named parameters in JS/TS that I'm aware of – named params in Scala would I suppose be an example of a more concise syntax.

jonathonherbert:
> Looks good! What’s the reason for changing the casing of `templateURL`? (I prefer the new casing, but just wondering if it’s for compatibility with something else)

Added note in PR description re: name change – it's what the iOS app already uses in production, so this saves them a build.

PR #44, opened 2024-06-26
Revert template URL work

jonathonherbert:
## What does this change?

Reverts #38 and #43 to remove the `templateURL` field, which requires additional work in iOS. 

## How to test

Deploy and reindex CODE. You should see images in recipes no longer contain a `templateURL` property.


Divs-B (approved):
Looks like needed as per iOS requirements 👍  thanks @jonathonherbert for mentioning that, Have we deployed on CODE and checked on response-data or any recipe on iOS device?

fredex42 (approved):
👍

jonathonherbert:
Tested in CODE, works as expected.

PR #45, opened 2024-07-09
sponsored recipes

Divs-B:
## What does this change?

We are testing 'sponsor' data inclusion in recipedata to parcel it to downstream

## How to test

unit tests is added

Tests are passing and can see sponsor data getting added in recipe data in the added testcase

## Accessibility

### DIFF
--- a/lib/recipes-data/src/lib/extract-recipes.test.ts
+++ b/lib/recipes-data/src/lib/extract-recipes.test.ts
@@ -1,7 +1,10 @@
 import {AssetType} from "@guardian/content-api-models/v1/assetType";
 import type {Block} from "@guardian/content-api-models/v1/block";
 import {ElementType} from "@guardian/content-api-models/v1/elementType";
+import type {Sponsorship} from "@guardian/content-api-models/v1/sponsorship";
+import {SponsorshipType} from "@guardian/content-api-models/v1/sponsorshipType";
 import {extractRecipeData} from "./extract-recipes";
+import {makeCapiDateTime} from './utils';
 
 jest.mock('./config', () => ({
   FeaturedImageWidth: 700,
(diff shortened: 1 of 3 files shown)


lib/recipes-data/src/lib/extract-recipes.test.ts
```diff
@@ -375,9 +378,123 @@ describe("extractRecipeData", () => {
...
+        },
+        validFrom: makeCapiDateTime("2023-09-21T16:18:10Z"),
+        validTo: makeCapiDateTime("2026-08-30T23:00:00Z"),
+      }
+    ]
+    const result = extractRecipeData(canonicalId, block, activeSponsorships)
+    expect(result.length).toEqual(1)
+    expect(result[0]?.recipeUID).toEqual("62ac3f0f98f6495cbefd72c11fac6d1e26390e99")
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-assignment -- this is safe
+    const data = JSON.parse(JSON.parse(JSON.stringify(result))[0]?.jsonBlob)
```
fredex42, line 494: Do we actually need to do `JSON.parse(JSON.stringify(result)[0]` ? Can we not just do `result[0]` ?

lib/recipes-data/src/lib/extract-recipes.test.ts
```diff
@@ -375,9 +378,123 @@ describe("extractRecipeData", () => {
...
+    const sponsorsExists = "sponsors" in data
+    expect(sponsorsExists).toBe(true)
```
fredex42, line 496: Great start - please can we verify that the fields actually come through correctly, and that there is exactly one sponsor in the data?

fredex42 (commented):
Great start - got a couple of requests on the tests

lib/recipes-data/src/lib/extract-recipes.test.ts
Divs-B: Sorry yes we can use result[0]. just changed 👍

lib/recipes-data/src/lib/extract-recipes.test.ts
Divs-B: yes now checking them as well 👍

emdash-ie (approved):

lib/recipes-data/src/lib/transform.ts
```diff
@@ -1,84 +1,109 @@
...
+export const addSponsorsTransform: (sponsors: Sponsorship[]) => RecipeTransformationFunction = sponsors => {
+  return (recipeData) => ({
+    ...recipeData,
+    sponsors: sponsors.length === 0 ? undefined : sponsors.map(sponsor => {
+      return {
+        ...sponsor,
+        sponsorshipType: getSponsorshipTypeNames[sponsor.sponsorshipType],
+        validFrom: sponsor.validFrom?.iso8601,
+        validTo: sponsor.validTo?.iso8601,
+        targeting: sponsor.targeting ? sponsor.targeting.publishedSince?.iso8601 : undefined
```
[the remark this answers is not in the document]
Divs-B: o yes ofcourse 😄 i missed putting same way as other two above using optional chaining. 
thanks for letting me know.

lib/recipes-data/src/lib/transform.ts
```diff
@@ -1,84 +1,109 @@
...
+  id: string;
+  featuredImage: RecipeImage | string; // the latter is an old image format that appears in our test fixtures
+  previewImage?: RecipeImage | string;
+};
+
+const getSponsorshipTypeNames: Record<number, string> = {
+  [SponsorshipType.SPONSORED]: "Sponsored",
+  [SponsorshipType.FOUNDATION]: "Foundation",
+  [SponsorshipType.PAID_CONTENT]: "Paid content"
 };
```
fredex42, line 51: It's a shame we can't get these strings out of the Thrift models :disappointed:

lib/recipes-data/src/lib/transform.ts
```diff
@@ -1,84 +1,109 @@
...
 
+export const addSponsorsTransform: (sponsors: Sponsorship[]) => RecipeTransformationFunction = sponsors => {
+  return (recipeData) => ({
+    ...recipeData,
+    sponsors: sponsors.length === 0 ? undefined : sponsors.map(sponsor => {
+      return {
+        ...sponsor,
+        sponsorshipType: getSponsorshipTypeNames[sponsor.sponsorshipType],
```
fredex42, line 59: We should deal with an "unknown" value here - `sponsorshipType` is a mandatory field in Thrift, so we should put something like `getSponsorshipTypeNames[sponsor.sponsorshipType] ?? "Unknown"` in case having an "undefined" value breaks the client json parsing

fredex42 (changes requested):
One little ask then I think we are ready to roll :grinning:

lib/recipes-data/src/lib/transform.ts
emdash-ie: I had a look through the typescript handbook, and I think we can, actually! Every enum apparently gets a [reverse mapping](https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings) from its values to the value names, which we could then postprocess ourselves to get the casing we want. The following seems to work as desired [in the typescript playround](https://www.typescriptlang.org/play/?ssl=35&ssc=1&pln=39&pc=1#code/KYOwrgtgBAygDgexAZwQJ2QCwJZwCoCecwUA3gFBRWwAKA8gHIx0BKAogCJQC8UADABpK1AGJ0Aqgw4BBPAElGPKAEYh1KDWlyOAfQDCjPGwZ4lAJnIBfcuQAuREgGlgBZEoDWLhADMo94j6wiCjoWLiExDYAxkioADbAAHRxCADmABT+wIHwsaE4+A4AlOQxIQnJaem5IRgFEcAA2jWodeEOiTD0TKycALol0bG2UKnAti357cQMAIYQwMgAXFAswDFoACYAPJNthcQCUMi2aNggqQB8ShRUzcGtYQdJXYzM7Bx9KwBEe8Cb3zUUHueX2DUSYkkMnkjC+UG+IgQYBAm1mtmwSEBwhBtSe4M02n0hmMeDh3xos2wmygZVsoFs3ysAG4bN5kVF0UhRuM5gt0icVns8cUViczhcyMI0OMwGgQEFQcLiI0Tn0rENykkUhkAAZjWy84DVbrvThFJQAElI+sN1QeU2enRNvQ4RUsOpKZXiWqqep58yNkKksgUDHNvCtNoDdsV9Q6QehobdHtKeQq2vSfoN0YJugMJhJ4agkf9fKFceIiVzRILJmTgzZIA5GPlNuwUWAtoFCtxFeARVFp3OqUlVGltll8vL0yaqsStgQABkEAB3YBoPSzZBGoos6ypzWVXVtju2149D5FksG9ud6PTx3n02u92etM+4882+2hMhxhX60v1Pe97TBeMJGDGEw1fA9vSPTMTzvPlq3zIw60tQCb2AstQKVJIUOJOsYPIRtmy5fUADV1wIBhv2jbsHwaAdjiHCVbhpYYoBSNc0CURiHBVWw+nnJdV3XTdt3SXcpRlOUoB1K1uPXRo+GEhdxDgYgNy3HdLEUsS0ESZA4lvdJlCKRJpTgOJZg7dJvh0QF4XhZN1VghB019SjqNo7CjSfF0AO8tAaLonDYxnJ03kCmCvQ8j8EPGKiQt8pDAwgxN-ww4LQr8mNe0i38oPrdzPM-WxktytL0gI2s8CCpKfLC-zcL7KstDzQj6pgoA):

```typescript
function getNiceName(st: SponsorshipType): string {
  const lower = SponsorshipType[st].toLowerCase();
  return `${lower[0].toUpperCase()}${lower.slice(1).replace("_", " ")}`
}
```

fredex42: 😁 @Divs-B would you mind having a go at @emdash-ie 's suggestion? I can pair with you if it gets tricky but I think it should just be a case of inserting the above function and replacing the lookup table with it?

the change should get picked up in the tests I hope.

lib/recipes-data/src/lib/transform.ts
fredex42: @Divs-B if you make the change above then this won't be necessary I hope

lib/recipes-data/src/lib/transform.ts
fredex42: (though we should deal with `SponsorshipType[st]` returning `undefined`, just in case the upstream models add a new value and we don't get updated)

Divs-B: Thanks @emdash-ie , I did try checking this by doing  `SponsorshipType[SponsorshipType[st]]` which is giving string literal of that, so yes reverse mapping definetly can be useful. I applied your code for string manipulation as well in exisitng code and included @fredex42 suggestion (below) for having "Unknown" sposnorship type if not found.

lib/recipes-data/src/lib/transform.ts
Divs-B: addressed in above change. thanks @fredex42

lib/recipes-data/src/lib/transform.ts
Divs-B: > 😁 @Divs-B would you mind having a go at @emdash-ie 's suggestion? I can pair with you if it gets tricky but I think it should just be a case of inserting the above function and replacing the lookup table with it?
> 
> the change should get picked up in the tests I hope.


Thanks @fredex42 for asking, its done now and should work.

fredex42 (approved):
Let's 🎢 it 😁

PR #47, opened 2024-07-17
Replace canonical article with CAPI ID

jonathonherbert:
## What does this change?

Adds a transformer to ensure that a recipe's `canonicalArticle` property always reflects the latest value from CAPI.

This property was previously added upstream in Composer – but it's not kept in sync, and there's no guarantee that the article path will not change after the element is added.

Better to derive it from fresh data.

## How to test

The automated test should pass, and it should be clear what it tests.

On `main`:

- Add a headline to a new article
- Add a recipe to an article (via the structuriser – manually adding is broken at the moment)
- Change the headline
- Note that in the recipe-backend API, the `canonicalArticle` property of the recipe has not updated.

Repeat on this branch. The `canonicalArticle` property should update.

- [x] Tested in CODE with this [article](https://composer.code.dev-gutools.co.uk/content/6697c1728f084da525701e30) – note that the recipe element has an incorrect path, but the recipe in [the CODE API](recipes.code.dev-guardianapis.com/content/FOJGV3iFruAHqxhq7jx-oRDkvYAzTP0lTEoiokB0yR0) has the correct path.

### DIFF
--- a/lib/recipes-data/src/lib/extract-recipes.test.ts
+++ b/lib/recipes-data/src/lib/extract-recipes.test.ts
@@ -5,6 +5,7 @@ import type {Sponsorship} from "@guardian/content-api-models/v1/sponsorship";
 import {SponsorshipType} from "@guardian/content-api-models/v1/sponsorshipType";
 import {extractRecipeData} from "./extract-recipes";
 import {makeCapiDateTime} from './utils';
+import { RecipeWithImageData } from "./transform";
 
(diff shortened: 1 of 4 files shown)


fredex42 (approved):
👍  Let's go

Divs-B (approved):
we did testing together and can see updated path. 
Looks great !! thanks @jonathonherbert

PR #49, opened 2024-07-25
Replace CAPI URL passed from Crier with channels-specific URL

jonathonherbert:
... so content solely launched to Feast can be retrieved.

## How to test

- The automated tests should pass. They've been adjusted for the new CAPI url.
- Modify [this article](https://composer.code.dev-gutools.co.uk/content/668503618f08a83a9aecabb3) and launch. You should see the change in the related recipe in the recipe backend content, e.g. [this recipe](https://recipes.code.dev-guardianapis.com/api/content/by-uid/3b3ba7fb30b24e83a6caef20eea83c91) (requires auth.) This would not occur prior to this PR.

Feast can publish recipes from articles that are large enough to trigger a RetrievableUpdate.

### DIFF
--- a/lambda/recipes-responder/src/update_retrievable_processor.test.ts
+++ b/lambda/recipes-responder/src/update_retrievable_processor.test.ts
@@ -18,7 +18,7 @@ jest.mock("./config", ()=>({
 }));
 
 const fakeUpdate:RetrievableContent = {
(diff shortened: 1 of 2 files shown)


fredex42 (approved):
😎 let's 🚢

PR #51, opened 2024-07-31
Publish status messages (read in Fronts tool)

jonathonherbert:
## What does this change?

Publishes status messages on curation deploy succeed or failure to an SNS topic, which is read in the Fronts tool.

## How to test

- The automated tests should pass.
- Running in CODE, attempt to publish a Feast issue in Fronts CODE. You should see confirmation of a publish in the event list.

Tested in CODE:

Screenshot 2024-08-02 at 13 52 29


jonathonherbert:
Currently failing to post messages to the SNS topic with 

```
Unhandled Promise Rejection 	{"errorType":"Runtime.UnhandledPromiseRejection","errorMessage":"Error: Resolved credential object is not valid"
```

cdk/lib/recipes-backend.ts
```diff
@@ -84,11 +84,33 @@ export class RecipesBackend extends GuStack {
...
-		const faciaSNSTopicARNParam = new GuParameter(this, 'faciaSNSTopicParam', {
-			default: `/${this.stage}/${this.stack}/${app}/facia-sns-topic-arn`,
-			fromSSM: true,
-			description: 'The ARN of the facia-tool SNS topic that emits curation notifications',
-		});
+	const faciaSNSTopicARNParam = new GuParameter(this, 'faciaSNSTopicParam', {
+		default: `/${this.stage}/${this.stack}/${app}/facia-sns-topic-arn`,
+		fromSSM: true,
+		description: 'The ARN of the facia-tool SNS topic that emits curation notifications',
+	});
```
emdash-ie, line 91: It looks like the difference in this paragraph is just the number of tabs used for indent. This file seems to use tabs and spaces for indentation in different places, and I suspect this indent is not actually what we want: to match the paragraphs above (which use spaces), we should probably use the existing indentation here, with a tab-width of 2 spaces.

Maybe settling on tabs or spaces in the file would be good though.

emdash-ie (approved):
Had [some trouble deploying to CODE](https://riffraff.gutools.co.uk/deployment/view/5de399f8-cbd3-4d9f-b5f5-5490cfb6d2f3), but other than that this looks good to me! One comment about the indentation in a file, but that’s pretty unimportant – can definitely merge without addressing that.

The trouble deploying to CODE is more troubling: it looks like it’s trying to recreate a dead-letter queue for some reason?

cdk/lib/recipes-backend.ts
jonathonherbert: It's interesting I didn't spot that – I've turned on whitespace rendering in my IDE, which should make it much clearer when VSCode is making interesting indentation choices! Thanks for spotting, [d927337](https://github.com/guardian/recipes-backend/pull/51/commits/d9273377dd62ec2250e3fa0e75760b549c9222d0)

emdash-ie: Configuring an automatic formatter is probably the easiest way to get consistency without ever having to worry about it: are we using prettier here? (Something to look at in a subsequent PR rather than this one)

jonathonherbert:
Thanks for the review! Previous deploy likely foiled by a typo fix in #53, which I've backported in [611a6c2](https://github.com/guardian/recipes-backend/pull/51/commits/611a6c29fad9e92c3c71205eda1636e9da07f403) so cloudformation doesn't recreate resources – I'll deploy  this shortly to check.

cdk/lib/recipes-backend.ts
jonathonherbert: Strong advocate of automatic linting, and have mentioned it in previous PR reviews on this repo, but shamefully never got around to implementing myself! This is a good cue to bump it up the priority list, agreed.

jonathonherbert:
Deployed now – recommend testing with an empty issue for now, e.g. https://fronts.code.dev-gutools.co.uk/v2/issues/10415d60-87c0-4f26-a6fa-9bcc5aa246aa, as Chef validation is waiting on https://github.com/guardian/facia-tool/pull/1612.

PR #53, opened 2024-08-01
Announce new recipes to EventBridge

fredex42:
## What does this change?

Outputs contents of new recipes to the eventbus as well as takedown notifications for recipes.

## How to test

- Create an SQS queue for testing
- Add a Rule to the `crier-eventbus-content-api-crier-v2-CODE` event bus with the following subscription, and link it to your SQS queue:
```json
{
  "source": ["recipe-responder"]
}
```
- (these two steps have been done already, going to the queue `test-recipes`)
- Go to the SQS console and open your queue
- Publish a recipe in CODE, using Composer (publishing only to the Feast channel will work too 😁 )
- In the SQS console, poll for new messages
- You should see messages coming through, indicating that a recipe was updated
- Perform a takedown on your recipe (or channel takedown)
- You should see another message indicating that the recipe was removed

Able to update the search index automatically

n/a.  We have discussed potential issues from clustering a lot of messages on one bus, but concluded that since the buses are stateless and we are not throwing around millions of messages per second it should be fine.

### DIFF
--- a/lambda/recipes-responder/src/update_processor.test.ts
+++ b/lambda/recipes-responder/src/update_processor.test.ts
@@ -5,7 +5,7 @@ import type {
   RecipeReferenceWithoutChecksum
 } from "@recipes-api/lib/recipes-data";
 import {
-  announce_new_recipe,
+  announceNewRecipe,
   calculateChecksum,
   extractAllRecipesFromArticle,
   insertNewRecipe,
@@ -25,7 +25,7 @@ jest.mock("@recipes-api/lib/recipes-data", () => ({
   recipesToTakeDown: jest.fn(),
   removeRecipeVersion: jest.fn(),
   sendTelemetryEvent: jest.fn(),
-  announce_new_recipe: jest.fn(),
+  announceNewRecipe: jest.fn(),
 }));
 
 const fakeContent: Content = {
@@ -157,7 +157,7 @@ describe("update_processor.handleContentUpdate", () => {
     expect((sendTelemetryEvent as Mock).mock.calls[0][0]).toEqual("PublishedData");
 
     // @ts-ignore -- Typescript doesn't know that this is a mock
-    expect(announce_new_recipe.mock.calls.length).toEqual(1);
+    expect(announceNewRecipe.mock.calls.length).toEqual(1);
   });
 
   it("should ignore a piece of content that is not an article", async () => {
@@ -204,7 +204,7 @@ describe("update_processor.handleContentUpdate", () => {
     expect((sendTelemetryEvent as Mock).mock.calls.length).toEqual(0);
 
     // @ts-ignore -- Typescript doesn't know that this is a mock
-    expect(announce_new_recipe.mock.calls.length).toEqual(0);
+    expect(announceNewRecipe.mock.calls.length).toEqual(0);
   });
 
   it("should be fine if there is no recipe content", async () => {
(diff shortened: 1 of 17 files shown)


lib/recipes-data/src/lib/takedown.ts
```diff
@@ -61,6 +62,13 @@ export async function removeAllRecipesForArticle(canonicalArticleId: string): Pr
   const removedEntries = await removeAllRecipeIndexEntriesForArticle(canonicalArticleId);
   console.log(`Taken down article ${canonicalArticleId} had ${removedEntries.length} recipes in it which will also be removed`);
   await Promise.all(removedEntries.map(recep=>removeRecipeContent(recep.checksum, "hard")));
+
+  try {
+    await announce_new_recipe([], removedEntries)
+  } catch(err) {
+    console.error("Unable to announce takedowns");
```
emdash-ie, line 69: Here as well I wonder whether it might be good to log the specific error?

emdash-ie (approved):
This looks good to me! 

Locally I seem to be having some trouble getting my editor set up: lots of warnings about unresolved imports, etc. I can take a look at that at some point, but also I see some of these errors when I run `npm run lint` at the top level of the repository: would be good to get that working to be sure the linter is running correctly.

I haven’t really used NX before, but this does seem like quite a complicated project setup: is it just my unfamiliarity with it?

lambda/recipes-responder/src/update_processor.test.ts
```diff
@@ -5,6 +5,7 @@ import type {
   RecipeReferenceWithoutChecksum
 } from "@recipes-api/lib/recipes-data";
 import {
+  announce_new_recipe,
```
jonathonherbert, line 8: nit nit nit: looking at the imports below it, s/snake/camel/? 😀

lib/recipes-data/src/lib/eventbus.test.ts
```diff
@@ -0,0 +1,92 @@
+import {EventBridgeClient, PutEventsCommand} from "@aws-sdk/client-eventbridge";
+import {mockClient} from "aws-sdk-client-mock";
```
jonathonherbert, line 2: TIL!

lib/recipes-data/src/lib/eventbus.ts
```diff
@@ -0,0 +1,73 @@
+import * as process from "process";
+import {EventBridgeClient,PutEventsCommand} from "@aws-sdk/client-eventbridge";
+import {registerMetric} from "@recipes-api/cwmetrics";
+import {OutgoingEventBus} from "./config";
+import type { RecipeIndexEntry, RecipeReference } from './models';
+
+const ebClient = new EventBridgeClient({region: process.env["AWS_REGION"]});
+
+export async function announce_new_recipe(updated:RecipeReference[], removedList:RecipeIndexEntry[]) {
+  if(!OutgoingEventBus || OutgoingEventBus =="") {
```
jonathonherbert, line 10: Should this be failing at the point we extract the config, rather than here? `mandatoryParameter` exists for this reason, I think.

jonathonherbert (commented):
A few non-blocking comments before I test, although it looks like @emdash-ie might get there first!

jonathonherbert (approved):
Tested end-to-end with Composer and https://github.com/guardian/recipe-search-backend/pull/18, LGTM 👍

lib/recipes-data/src/lib/eventbus.ts
fredex42: These are here because event bus was not a mandatory parameter - the responder still works even if eventbus doesn't so we don't break publication (error thrown here is caught and non-fatal). Can change in a future PR but I'd rather keep as-is right now

PR #54, opened 2024-08-02
Better publish messages on Fronts parse error

jonathonherbert:
~NB: don't merge until we've merged https://github.com/guardian/facia-tool/pull/1612, which should fix any parsing issues!~ Now merged, this is ready for review.

## What does this change?

Better error messages sent to facia-tool's issue publication status queue when parsing the Fronts data fails.

To do this, we split the parsing into two steps – the first, for the message envelope containing the issue data that lets us post a notification containing the relevant information for facia-tool, and the second, parsing the Fronts data itself.

The Fronts data is much more likely to change over time, and it's nice to receive errors in the upstream tooling when downstream breaks.

## How to test

- The tests should pass. They test the happy path, an unparseable message, a failed envelope parse, and a failed fronts parse.
- Deploy a version of the Fronts tool prior to https://github.com/guardian/facia-tool/pull/1612, and publish a Feast issue. You should see a validation error appear in the publication status messages.

A few tab/space inconsistency issues are covered in facia-models.ts – perhaps better reviewed with the [whitespace diff off.](https://github.com/guardian/recipes-backend/pull/54/files?diff=split&w=1)

### DIFF
--- a/lambda/facia-responder/src/main.ts
+++ b/lambda/facia-responder/src/main.ts
@@ -2,13 +2,26 @@ import * as facia from '@recipes-api/lib/facia';
 import { deployCurationData } from '@recipes-api/lib/recipes-data';
 import type { SNSMessage, SQSHandler, SQSRecord } from 'aws-lambda';
 import format from 'date-fns/format';
+import type { SafeParseReturnType } from 'zod';
 import { notifyFaciaTool } from './facia-notifications';
 import { getErrorMessage } from './util';
 
-function parseMesssage(from: SQSRecord): facia.FeastCuration {
+function getMessageBodyAsObject(from: SQSRecord): unknown {
 	const parsedSNSMessage = JSON.parse(from.body) as SNSMessage; // will throw if the content is not valid;
 	const parsedBody = JSON.parse(parsedSNSMessage.Message) as unknown;
-	return facia.FeastCuration.parse(parsedBody);
+	return parsedBody;
+}
+
+function parseFeastCurationEnvelope(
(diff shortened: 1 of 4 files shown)


lib/facia/src/lib/facia-models.ts
```diff
@@ -2,60 +2,60 @@ import { z } from "zod";
 
 //This should mirror FeastAppModel in Facia backend
 export const RecipeIdentifier = z.object({
-  id: z.string()
+	id: z.string()
```
jonathonherbert, line 5: spaces -> tabs, consistent w/ rest of project

lambda/facia-responder/src/main.ts
```diff
@@ -37,28 +50,58 @@ export const handler: SQSHandler = async (event) => {
...
+			return notifyFaciaTool({
+				edition,
+				issueDate,
+				version,
+				status: 'Published',
+				message: 'This issue has been published',
+				timestamp: Date.now(),
+			});
 		} catch (e) {
 			return notifyFaciaTool({
```
fredex42, line 97: Do we want to log the error out here too do you think?

fredex42 (approved):
Great! One little question but 👍 for it to go

lambda/facia-responder/src/main.ts
jonathonherbert: Good shout! [701dc99](https://github.com/guardian/recipes-backend/pull/54/commits/701dc99798803616bc54044646c03aeaacb23e8b)

PR #55, opened 2024-08-02
Make chef bio optional

jonathonherbert:
## What does this change?

Makes the chef bio optional. We'll want to pull in bio information from the tag page directly (as with images) in the curation lambda if there are no overrides.

## How to test

In Fronts CODE with https://github.com/guardian/facia-tool/pull/1612 deployed, add a chef to an issue, and publish. You should see the publish request succeed:

Screenshot 2024-08-02 at 14 40 25

Previous requests would fail with a Zod error ([example logs.](https://logs.gutools.co.uk/s/content-platforms/app/discover#/doc/b0be43a0-59d7-11e8-a75a-b7af20e8f748/logstash-capi-2024.08.02?id=_D9LE5EBb0N6Anqergf0))

### DIFF
--- a/lib/facia/src/lib/facia-models.ts
+++ b/lib/facia/src/lib/facia-models.ts
@@ -17,7 +17,7 @@ export const Chef = z.object({
   backgroundHex: z.string().optional(),
   id: z.string(),
   image: z.string().optional(),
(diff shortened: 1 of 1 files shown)


emdash-ie (approved):
Haven’t tested in CODE, but this looks fine.

PR #56, opened 2024-08-08
Document use of API Gateway keys in readme

emdash-ie:
I was unaware recently of how the auth worked, and this note would have helped me find how to create a new API key more quickly.


codingWithRach (approved):
Looks helpful 👍

PR #58, opened 2024-08-12
bump library versions

fredex42:
## What does this change?

Bumps library versions

## How to test

Deploy to CODE, publish a recipe from Composer CODE, monitor logs and check it is present on the recipe backend API


Divs-B (approved):
Good to go, thanks Andy for upgrading and testing 👍

PR #59, opened 2024-08-15
To exclude sponsorship in older app version

Divs-B:
## What does this change?

Co-authored by: @fredex42 

Details in trello card: https://trello.com/c/5UNtyAua/3105-how-do-we-exclude-sponsored-content-from-older-app-versions

We have added `sponsorshipCount` to current recipe structure so that based on the count (0 means not Sponsored and sponsored otherwise), recipe can make it to newer version of apps otherwise not to older version of apps.

## How to test

-Deploy to CODE 
-Publish any existing recipe or create new with tesco tag attached to it (means make it sponsored),
-Observe logs - it should reflect path getting updated if sponsored then we should see something like `updating s3://recipes-backend-static-code/v2/index.json` otherwise to `index.json`.
-Can check by hitting `https://recipes.code.dev-guardianapis.com/index.json` to see sponsired article should not presrnt in that and on `https://recipes.code.dev-guardianapis.com/v2//index.json` to see presence of your sponsored recipe .

Unblock the progress towards sponsored app content


lambda/recipes-responder/src/constants.ts
```diff
@@ -0,0 +1,4 @@
+//todo - as of now creating for recipes-responder but we can check if we need constant file globally means at root of the project to hold all constants?)
+
+export const INDEX_JSON = "index.json"
+export const V2_INDEX_JSON = "v2/index.json"
```
fredex42, line 4: These need to go into the common library as they are currently breaking module scoping rules by being imported in writeIndexData which lives in the common library

lambda/recipes-responder/src/commandline-reindex.ts
```diff
@@ -150,15 +160,16 @@ async function main() {
 
   console.log("------------------------------------------------------");
   console.log("Rebuilding index...");
-  const indexData = await retrieveIndexData();
-  await writeIndexData(indexData);
+  const indexDataForAllRecipes = await retrieveIndexData();
+  await indexing(indexDataForAllRecipes, test)
+
```
fredex42, line 165: ☝️ This needs to be updated to handle the split index - I messed around with it while testing and haven't fixed it up, so will do that next

lambda/recipes-responder/src/main.ts
```diff
@@ -1,19 +1,19 @@
 import {EventType} from "@guardian/content-api-models/crier/event/v1/eventType";
 import {ItemType} from "@guardian/content-api-models/crier/event/v1/itemType";
-//import type {KinesisStreamHandler, KinesisStreamRecord} from "aws-lambda";
 import type {EventBridgeHandler} from "aws-lambda"
 import {registerMetric} from "@recipes-api/cwmetrics";
 import {deserializeEvent} from "@recipes-api/lib/capi";
 import {retrieveIndexData, writeIndexData} from "@recipes-api/lib/recipes-data";
+import {INDEX_JSON, V2_INDEX_JSON} from "./constants"
```
fredex42, line 7: These can be imported from the common library (see my earlier point) but must be re-exported from the `index.ts` of the library to do that

lambda/test-indexbuild/src/main.ts
```diff
@@ -1,23 +1,21 @@
...
-import { PutObjectCommand, S3Client } from "@aws-sdk/client-s3";
-// eslint-disable-next-line import/no-unresolved -- this works from root tsconfig but eslint keeps complaining :-(
-import { type RecipeIndex, retrieveIndexData } from "@recipes-api/lib/recipes-data";
-import type { Handler } from "aws-lambda";
-import { StaticBucketName as Bucket } from "./config";
-
+import {PutObjectCommand, S3Client} from "@aws-sdk/client-s3";
+import type {Handler} from "aws-lambda";
+import {type RecipeIndex, retrieveIndexData} from "@recipes-api/lib/recipes-data";
+import {INDEX_JSON, V2_INDEX_JSON} from "../../recipes-responder/src/constants";
```
fredex42, line 5: This is the line that triggers a module-boundary warning.  As I mentioned before, if you want to share a value across multiple projects in this way you need to put it into the common library (which is fairly easy 😀 )

lambda/test-indexbuild/src/main.ts
```diff
@@ -27,12 +25,38 @@ async function writeIndexData(indexData:RecipeIndex)
...
+    if(!indexDataForUnSponsoredRecipes.recipes.find(r=>r.capiArticleId===entry.capiArticleId)) {
+      console.log(entry);
+    }
+  }
+
+  const test = indexDataForUnSponsoredRecipes.recipes.find(r=>r.capiArticleId==="lifeandstyle/article/2024/aug/19/testing-sponsor-article-for-v2");
+  console.log("unsponsored: ", test);
+
+  const test2 = indexDataForAllRecipes.recipes.find(r=>r.capiArticleId==="lifeandstyle/article/2024/aug/19/testing-sponsor-article-for-v2");
+  console.log("sponsored: ", test2);
```
fredex42, line 51: Can tidy up this debug - I'm happy to do that, I made the mess in the first place 😁

fredex42 (commented):
Nice work @Divs-B  - there are a few things for us both to clear up, I'll get on with mine posthaste!

fredex42:
By the way... @Divs-B don't merge this immediately.  The deploy will fail due to DynamoDB being annoying and not allowing you to make the requisite changes to global secondary indexes in one hit.  When we deploy, we'll need to handle the data migration in the following way:

- temporarily disable the recipe-responder lambda function
- run a local build of the CDK with the relevent global secondary index setup commented out.  Deploy this manually via CFN console
- this will delete the GSI
- _then_ merge and deploy.  The GSI will be re-created under the same name
- then wait a couple of minutes while the index is re-populated
- then un-disable the recipe-responder

Divs-B:
> By the way... @Divs-B don't merge this immediately. The deploy will fail due to DynamoDB being annoying and not allowing you to make the requisite changes to global secondary indexes in one hit........

Thanks @fredex42 for all updates and also making changes for fixing efficiency and reading records from dynamo in cdk as well. 
Will try to pair on merging process for precaution.

lambda/recipes-responder/src/constants.ts
Divs-B: Thanks @fredex42 . constants file moved to common folder like writeIndex as suggested and hence solves todo question too to keep constants file here or somewhere else.

fredex42 (approved):
🚢

PR #60, opened 2024-08-22
Correctly validate mise-en-place curation data

fredex42:
## What does this change?

Updates our internal data models to ensure that real data from the current version of MEP can be validated

## How to test

- Unit-testing is provided against a front downloaded on 22nd Aug 2024
- To test "in the wild", deploy to CODE
- Monitor the logs in ELK using the key `app.keyword: recipes-backend-rest-endpoints`
- Filter to `WARN` logs (or filter _out_ `DEBUG` and `INFO`)
- Upload a curation file from mise-en-place to the POST endpoint.
- You should no longer see the `WARN` logs about the format not being correct (if you expand the ELK view to 30 days you should see plenty of them)

Able to manipulate the data, so we can apply filtering

- The Facia models will have to be updated to correspond to these

### DIFF
--- a/lib/facia/src/lib/facia-models.test.ts
+++ b/lib/facia/src/lib/facia-models.test.ts
@@ -1,8 +1,18 @@
-import type { Recipe} from './facia-models';
-import {FeastCuration} from './facia-models';
+import * as fs from "node:fs";
+import * as path from "path";
+import type {Chef, Recipe} from './facia-models';
+import {FeastCuration,MiseEnPlaceData} from './facia-models';
+
 
 describe("facia-models", ()=>{
-  it("should be able to validate real data", ()=>{
+  function loadFixture(name:string) {
+    const filepath = path.join(__dirname, name);
+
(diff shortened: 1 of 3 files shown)


jonathonherbert (approved):
👍 

Re: `MiseEnPlaceDataFormat` etc.: are we planning on migrating MEP data to FeastCuration format? If not, is this type not tool-agnostic, as Feast Fronts will be writing the same format?

fredex42:
> 👍
> 
> Re: `MiseEnPlaceDataFormat` etc.: are we planning on migrating MEP data to FeastCuration format? If not, is this type not tool-agnostic, as Feast Fronts will be writing the same format?

Technically - `MiseEnPlaceDataFormat` is just the _wrapper_ around the actual curation data (single-front array of collections) as opposed to `FeastCurationData` which is Fronts tool's object wrapper (giving the edition and multiple fronts, which are then the same array of collections).

When we move over to Fronts, `MiseEnPlaceDataFormat` is deprecated and can be removed

PR #61, opened 2024-08-22
Adds function to return boolean indicating whether recipe is sponsored

codingWithRach:
## What does this change?

Changes the `idxRecipeUID` index on the `recipes-backend-indexstore-STAGE` table so that it returns all fields instead of just returning the recipeVersion

Creates a function `isSponsored` which accepts the recipeUID, returning True if the `sponsorshipCount` field exists in the corresponding DynamoDB record and is true; otherwise False

If the record cannot be found or an error occurs during the retrieval, an error is thrown so that conclusions are not incorrectly drawn about whether or not the recipe is sponsored

## How to test

Run `npm test` 

## Accessibility

### DIFF
--- a/lambda/publish-todays-curation/src/is-sponsored.ts
+++ b/lambda/publish-todays-curation/src/is-sponsored.ts
@@ -1,0 +1,36 @@
+import { DynamoDBClient, QueryCommand } from '@aws-sdk/client-dynamodb';
+import { Stack, Stage } from './config';
+
+interface RecipeItem {
+	sponsorshipCount?: { N: string };
+}
+
+export async function isSponsored(recipeUID: string): Promise<boolean> {
+	const maybePreview = Stack.endsWith('-preview') ? '-preview' : '';
+	const tableName = `recipes-backend${maybePreview}-indexstore-${Stage}`;
+	try {
+		const client = new DynamoDBClient({ region: process.env.AWS_REGION });
+		const req = new QueryCommand({
+			TableName: tableName,
+			IndexName: 'idxRecipeUID',
+			KeyConditionExpression: 'recipeUID = :uid',
+			ExpressionAttributeValues: {
+				':uid': { S: recipeUID },
+			},
+		});
+
+		const response = await client.send(req);
+		if (response.Items && response.Items.length > 0) {
+			const item = response.Items[0] as RecipeItem;
+			if (item.sponsorshipCount?.N && parseInt(item.sponsorshipCount.N) > 0) {
+				return true;
+			}
+		}
+	} catch (error) {
+		console.log(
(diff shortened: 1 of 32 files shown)


lambda/publish-todays-curation/src/is-sponsored.ts
```diff
@@ -0,0 +1,41 @@
...
+		if (response.Items && response.Items.length > 0) {
+			const item = response.Items[0] as RecipeItem;
+			return (
+				!!item.sponsorshipCount?.N && parseInt(item.sponsorshipCount.N) > 0
+			);
+		} else {
+			console.log(
+				`ERROR [${recipeUID}] - valid recipe not found in ${tableName}`,
+			);
+			return false;
```
fredex42, line 32: Hmmmmm...... good point.  I _would_ be tempted to throw an exception here - because it's not true that it either IS sponsored or NOT sponsored if it does not exist, and it's not clear what the caller should do in that case.  So I'd make it an explicit third state.

lambda/publish-todays-curation/src/is-sponsored.ts
```diff
@@ -0,0 +1,41 @@
...
+		}
+	} catch (error) {
+		console.log(
+			`ERROR [${recipeUID}] - error retrieving recipe from ${tableName}`,
+			error,
+		);
+		return false;
```
fredex42, line 39: Ditto - in fact, if we are allowing the function to throw, we could just remove the try/catch entirely and rely on the caller to catch it.

fredex42 (commented):
Great! Couple of small points inline - on balance I think we _should_ throw an exception here

lambda/publish-todays-curation/src/is-sponsored.ts
```diff
@@ -0,0 +1,41 @@
+import { DynamoDBClient, QueryCommand } from '@aws-sdk/client-dynamodb';
+import { Stack, Stage } from './config';
+
+interface RecipeItem {
+	sponsorshipCount?: { N: string };
+}
+
+export async function isSponsored(recipeUID: string): Promise<boolean> {
+	const maybePreview = Stack.endsWith('-preview') ? '-preview' : '';
+	const tableName = `recipes-backend${maybePreview}-indexstore-${Stage}`;
```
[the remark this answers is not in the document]
codingWithRach: Ah the joy of an unfamiliar codebase is being able to reinvent the wheel (unnecessarily)!

fredex42:
We are putting this on hold for the moment..... Feast are currently OK to proceed with the way things are. Adding `do-not-merge` but keeping it around in case it suddenly turns out to be needed!

PR #63, opened 2024-09-13
Insert article time data into recipe model

codingWithRach:
## What does this change?

Extracts time data from the article and inserts into the recipe model, so that it can be used in the recipe search.  

It's currently unclear which time data will be most useful and so, for now, the following fields are being extracted:
- lastModifiedDate
- firstPublishedDate
- publishedDate

## How to test

Run `npm test` - tests pass

Deploy to CODE
Find a recipe article on the Guardian website and use `Teleporter | flexible composer | Management | View previous versions` to `Restore` the recipe to CODE
Access `recipe-responder` in ElasticSearch and filter to CODE - this will check it hasn't broken
Locate the DEBUG message with `urlToPurge` that does **not** relate to index.json, and extract the part of the url from `recipes.code` onwards
Append this to `https://` in a new browser window
View the output and check that the following have been populated:
- lastModifiedDate
- firstPublishedDate
- publishedDate

## Accessibility

### DIFF
--- a/lib/recipes-data/src/lib/extract-recipes.test.ts
+++ b/lib/recipes-data/src/lib/extract-recipes.test.ts
@@ -7,7 +7,6 @@ import Int64 from 'node-int64';
 import { extractRecipeData } from './extract-recipes';
 import type { RecipeWithImageData } from './transform';
 import { capiDateTimeToDate, makeCapiDateTime } from './utils';
-import { RecipeDates } from './models';
 
 jest.mock('./config', () => ({
 	FeaturedImageWidth: 700,
--- a/lib/recipes-data/src/lib/extract-recipes.test.ts
+++ b/lib/recipes-data/src/lib/extract-recipes.test.ts
@@ -1,6 +1,6 @@
 import type { Sponsorship } from '@guardian/content-api-models/v1/sponsorship';
 import { extractRecipeData } from './extract-recipes';
-import { activeSponsorships, block, canonicalId, invalidRecipeElements, multipleRecipeElements, noRecipeElements, recipeDates, singleRecipeElement } from './recipe-fixtures';
+import { activeSponsorships, block, content, invalidRecipeElements, multipleRecipeElements, noRecipeElements, recipeDates, singleRecipeElement } from './recipe-fixtures';
 import type { RecipeWithImageData } from './transform';
 import { capiDateTimeToDate } from './utils';
 
@@ -13,7 +13,7 @@ jest.mock('./config', () => ({
 describe('extractRecipeData', () => {
 	it('should work when block containing single recipe element', () => {
 		const testBlock = { ...block, elements: singleRecipeElement };
-		const result = extractRecipeData(canonicalId, testBlock, []);
+		const result = extractRecipeData(content, testBlock, []);
 		expect(result.length).toEqual(1);
 		expect(result[0]?.recipeUID).toEqual(
 			'62ac3f0f98f6495cbefd72c11fac6d1e26390e99',
@@ -34,7 +34,7 @@ describe('extractRecipeData', () => {
 
 	it('should work when block containing multiple recipe elements', () => {
 		const testBlock = { ...block, elements: multipleRecipeElements}
-		const result = extractRecipeData(canonicalId, testBlock, []);
+		const result = extractRecipeData(content, testBlock, []);
 		expect(result.length).toEqual(3);
 		expect(result[2]?.recipeUID).toEqual(
 			'62ac3f0f98f6495cbefd72c11fac6d1e26390e99',
@@ -55,20 +55,20 @@ describe('extractRecipeData', () => {
 
 	it('should work when block containing no recipe elements ', () => {
 		const testBlock = { ...block, elements: noRecipeElements}
-		const result = extractRecipeData(canonicalId, testBlock, []);
+		const result = extractRecipeData(content, testBlock, []);
 		expect(result.length).toEqual(0);
 	});
 
 	it('should return empty array when block has got invalid recipe element (no ID field) ', () => {
 		const testBlock = { ...block, elements: invalidRecipeElements}
-		const recipesFound = extractRecipeData(canonicalId, testBlock, []);
+		const recipesFound = extractRecipeData(content, testBlock, []);
 		expect(recipesFound).toEqual([null]);
 		expect(recipesFound.length).toEqual(1);
 	});
 
 	it('should work when block containing recipe element and is sponsored, means we have sponsorship data as well', () => {
 		const testBlock = { ...block, elements: singleRecipeElement };
-		const result = extractRecipeData(canonicalId, testBlock, activeSponsorships);
+		const result = extractRecipeData(content, testBlock, activeSponsorships);
 		expect(result.length).toEqual(1);
 		expect(result[0]?.recipeUID).toEqual(
 			'62ac3f0f98f6495cbefd72c11fac6d1e26390e99',
@@ -88,7 +88,7 @@ describe('extractRecipeData', () => {
 	it('should work when block containing recipe element but not sponsored, means no sponsor data available', () => {
 		const testBlock = { ...block, elements: singleRecipeElement };
 		const activeSponsorships: Sponsorship[] = [];
-		const result = extractRecipeData(canonicalId, testBlock, activeSponsorships);
+		const result = extractRecipeData(content, testBlock, activeSponsorships);
 		expect(result.length).toEqual(1);
 		expect(result[0]?.recipeUID).toEqual(
 			'62ac3f0f98f6495cbefd72c11fac6d1e26390e99',
@@ -103,8 +103,9 @@ describe('extractRecipeData', () => {
 
 	it('should update the canonicalArticle with the content ID', () => {
 		const canonicalId = 'a-new-path';
+    const testContent = { ...content, id: canonicalId };
     const testBlock = { ...block, elements: singleRecipeElement };
-		const result = extractRecipeData(canonicalId, testBlock, []);
+		const result = extractRecipeData(testContent, testBlock, []);
 		expect(result.length).toEqual(1);
 		const data = JSON.parse(
 			result[0]?.jsonBlob as string,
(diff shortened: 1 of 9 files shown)


lib/recipes-data/src/lib/extract-recipes.test.ts
emdash-ie, line 1: The formatting changes in this file made it slightly tricky to see the actual changes – it would’ve been handy to split them out into a separate commit, though I know that can be awkward.

Also, this file is now pretty big, with a good bit of repetition: maybe we could reduce some of it by introducing common variables for some of it? (E.g., the recipe json, the whole Block, or the html paragraphs?) Not something to do in this PR, but it might make our lives easier.

emdash-ie (approved):
Looks good! Would be good to see the results of testing in CODE before merging though

lib/recipes-data/src/lib/extract-recipes.test.ts
fredex42: Yeah I vote we split the fixtures out into a seperate file and import them from there

lib/recipes-data/src/lib/transform.ts
```diff
@@ -69,6 +69,15 @@ export const addSponsorsTransform: (sponsors: Sponsorship[]) => RecipeTransforma
   })
 }
 
+export const addRecipeDatesTransform: (recipeDates: RecipeDates) => RecipeTransformationFunction = recipeDates => {
+  return (recipeData) => ({
+    ...recipeData,
+    ...(recipeDates.lastModifiedDate && { lastModifiedDate: recipeDates.lastModifiedDate.toISOString() }),
+    ...(recipeDates.firstPublishedDate && { firstPublishedDate: recipeDates.firstPublishedDate.toISOString() }),
+    ...(recipeDates.publishedDate && { publishedDate: recipeDates.publishedDate.toISOString() }),
```
fredex42, line 77: Little bit niggly as I found this hard to read, probably be easier to simply put it as:

```typescript
return (recipeDate) => ({
   ...recipeData,
   lastModifiedDate: recipeDates.lastModifiedDate ? recipeDates.lastModifiedDate.toISOString() : undefined,
   firstPublishedDate: recipeDates.firstPublishedDate ? recipeDates.firstPublishedDate.toISOString() : undefined,
```
etc/

fredex42 (changes requested):
Great job, got one ask in the code and as @emdash-ie the test file got a bit tricky - would you mind separating the fixture data into a seperate .ts file from where it can be imported?

lib/recipes-data/src/lib/extract-recipes.test.ts
fredex42: Linting is a bit annoying - I think we need to get eslint/prettier working properly on this so that all editors apply formatting consistently.

codingWithRach: I've attempted to address the eslint/prettier issue in PR #65

lib/recipes-data/src/lib/utils.ts
```diff
@@ -30,6 +30,13 @@ export function makeCapiDateTime(from: string): CapiDateTime {
   }
 }
 
+export function capiDateTimeToDate(date: CapiDateTime | undefined): Date | undefined {
+  if (!date) {
+    return undefined;
+  }
+  return new Date(date.iso8601);
+}
```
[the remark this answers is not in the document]
codingWithRach: Absolutely - much neater!

fredex42 (approved):
😁  Great! Let's run it up on CODE and check it before merging

lib/recipes-data/src/lib/extract-recipes.ts
```diff
@@ -37,16 +39,32 @@ export async function extractAllRecipesFromArticle(content: Content): Promise<Re
 }
 
 
-export function extractRecipeData(canonicalId: string, block: Block, sponsorship: Sponsorship[]): Array<RecipeReferenceWithoutChecksum | null> {
+export function extractRecipeData(content: Content, block: Block, sponsorship: Sponsorship[]): Array<RecipeReferenceWithoutChecksum | null> {
```
Divs-B, line 42: If we have decided to pass whole `content` as parameter then do we need `block` as separate argument in this method call?
If i can see correctly then `block` is `content.blocks.main`?

Divs-B (changes requested):

lib/recipes-data/src/lib/extract-recipes.ts
codingWithRach: No - extractRecipeData is called from several places.  The first time it passes `articleBlocks.main` but on subsequent occasions it does a flatMap through `articleBlocks.body`

Divs-B: Yes, it does flatmap and iterate through different blocks which was this method's original goal for,  but i think passing whole `content` and then its internal data `block` both are sounding confusing technically. Either we should be able to pass `publicationDate` as parameter to use further for `content.fields?.firstPublicationDate` to construct `recipeDates`?

Divs-B (changes requested):

lib/recipes-data/src/lib/extract-recipes.ts
Divs-B: Yes, it does flatmap and iterate through different blocks which was this method's original goal for, but i think passing whole content and then its internal data block both are sounding confusing technically. 
Either we should be able to pass then `publicationDate` as parameter to use further for content.fields?.firstPublicationDate to construct recipeDates?

codingWithRach: I see your point, but on the other hand determining the firstPublicationDate and publicationDate before calling extractRecipeData would result in extra processing as I currently only need to find the fallback options if the main ones don't exist (whereas your suggestion would mean having to calculate the fallback options whether they're needed or not).  So pros and cons.

JustinPinner:
Theres a lot of conversation here already so feel free to ignore this, but how about using the latest of `block.createdDate` and `block.lastModifiedDate`?  I don't know if we'd need a fallback from there, but I think blocks always have at least a created date (TBC).

lib/recipes-data/src/lib/extract-recipes.ts
fredex42: Hmmmmmm, I see both points. I can imagine one way of doing it that's a bit different, more scala-y, that might work.

Something like this:

```typescript
interface RecipeDateTime {
  lastModified:Date;
  firstPublished:Date;
  published:Date;
}

function RecipeDateTimeFromCapi(lastModified:CapiDateTime, firstPublished:CapiDateTime, published:CapiDateTime): RecipeDateTime {
  return {
    lastModified: new Date(lastModified.dateTime.toNumber()),
    firstPublished: new Date(firstPublished.dateTime.toNumber()),
    published: new Date(published.dateTime.toNumber())
  }
}

function UpdateRecipeDateTimeFromBlock(block:Block, source:RecipeDateTime):RecipeDateTime {
  return {
    lastModified:  block.lastModifiedDate ? new Date(block.lastModifiedDate.dateTime.toNumber()) : source.lastModified,
    firstPublished: block.firstPublishedDate ? new Date(block.firstPublishedDate.dateTime.toNumber()) : source.firstPublished,
    published: block.publishedDate ? new Date(block.publishedDate.dateTime.toNumber()): source.published
  }
}
```

Then, we can call like this:

```typescript
const recipeDates = RecipeDateTimeFromCapi(block.lastModifiedDate, block.firstPublishedDate, block.publishedDate);
const getAllMainBlockRecipesIfPresent = extractRecipeData(content.id, articleBlocks.main as Block, rdt, sponsorship);

.
.
.
if (!block?.elements) return [];
  else {
   const finalRecipeDates = UpdateRecipeDateTimeFromBlock(block, recipeDates);
    return block.elements
      .filter(elem => elem.type === ElementType.RECIPE)
      .map(recp => parseJsonBlob(content.id, recp.recipeTypeData?.recipeJson as string, sponsorship, finalRecipeDates))
  }
}
```

what do you think?

Divs-B: This looks really good, not confusing as well, i think we can try this if @codingWithRach also agrees.

Just one quick check: Unless I missed reading it properly Is this code taking both places `block` dates for block and source?, i think we need to pass `content`'s field's publicationdate at one place to get as source dates and then compare if block's dates are available or not otherwise contnet's dates. Let me know if not this.

codingWithRach: Agreed that that looks like a neater solution, although @Divs-B is correct that one of them needs to be populated from the section rather than the block.  I'll have a look at it now!

codingWithRach: So here's a suggestion (although it does involve calculating the fallback values whether they're required or not):

**extractAllRecipesFromArticle**

```
    const fallbackRecipeDates: RecipeDates = {
      lastModifiedDate: undefined,
      firstPublishedDate: getFallbackFirstPublishedDate(content),
      publishedDate: getFallbackPublishedDate(content)
      };
```

**extractRecipeData**

`const recipeDates: RecipeDates = updateRecipeDatesFromBlock(block, fallbackRecipeDates)`

**New functions**

```
export function getFallbackFirstPublishedDate(content: Content): Date | undefined {
  return content.fields?.firstPublicationDate ? capiDateTimeToDate(content.fields.firstPublicationDate) : undefined;
}

export function getFallbackPublishedDate(content: Content): Date | undefined {
  const feastChannel = content.channels?.find(channel => channel.channelId === 'feast');
  return feastChannel?.fields.publicationDate ? capiDateTimeToDate(feastChannel.fields.publicationDate) : undefined;
}

export function updateRecipeDatesFromBlock(block: Block, fallbackRecipeDates: RecipeDates): RecipeDates {
  return {
    lastModifiedDate: block.lastModifiedDate ? capiDateTimeToDate(block.lastModifiedDate): fallbackRecipeDates.lastModifiedDate,
    firstPublishedDate: block.firstPublishedDate ? capiDateTimeToDate(block.firstPublishedDate): fallbackRecipeDates.firstPublishedDate,
    publishedDate: block.publishedDate ? capiDateTimeToDate(block.publishedDate): fallbackRecipeDates.publishedDate,
  }
}
```

Divs-B: Thanks @codingWithRach for looking into this and refining. It will be great to have refactoring in follow up PR. Let this go ahead 👍

Divs-B (approved):

lib/recipes-data/src/lib/extract-recipes.ts
fredex42: Yeah let's refactor it going forwards

fredex42 (approved):
🚢

PR #64, opened 2024-09-16
Endpoint for batch resolve uids

fredex42:
## What does this change?

Adds a new endpoint that resolves recipe immutable IDs to index entries in bulk

**GET** /api/content/by-uid?ids=id-1,id-2,id-3,id-4

### Note

You should check the Fastly CDN settings when deploying this, as there are a couple of bugs which can cause the GET parameters on the request to be lost in transit.

1. Under "Hosts", the static backend has a condition called `IF Is not API`. The first part of this must read `! (req.url.path ~ "^/prod/api" `, NOT `! (req.url.path ~ "^/api" `
2. Under "VCL Snippets", the snipped called `Protected S3 Access` might need editing.  After the variable setup is an `if` statement which acts as a "master switch".  This must contain the directive ` && !(req.url.path ~ "/api")` in order for the parameters to be passed correctly. Note that it is `/api` NOT `/prod/api` as it is applied higher in the configuration than the rewrite for api path

## How to test

- Deploy to CODE and check CDN settings as above
- Grab some valid UUIDs from the CODE environment: 
```bash
curl https://recipes.code.dev-guardianapis.com/v2/index.json | jq -r '.recipes[].recipeUID' | head -n 5
```
- Build a GET request to the endpoint as above
- Remember to send your recipes API key in the `X-Api-Key` header. You can obtain this from the "recipes-backend-CODE" Usage Plan in the API Gateway console
- You should see a result like this:
```json
{
    "status": "ok",
    "resolved": 3,
    "requested": 3,
    "results": [
        {
            "checksum": "xyKmLw49Ro-tKGTioDqN-h4a_vgmbkchcvKq_hmoB8A",
            "recipeUID": "3f3410b0edae470abb76f5f48c02e8f4",
            "capiArticleId": "food/2021/aug/23/thomasina-miers-recipe-summer-tomato-tagliatelle-toasted-walnut-nduja-pesto",
            "sponsorshipCount": 0
        },
        {
            "checksum": "eKcNovgTsLfiXN9YLa13mXa3cs4O4oc5mg0YNbmMvKE",
            "recipeUID": "d90776464018f0515e6cf979790e446d14588413",
            "capiArticleId": "lifeandstyle/wordofmouth/2011/jan/20/how-make-perfect-orange-marmalade",
            "sponsorshipCount": 0
        },
        {
            "checksum": "5FoOLlNf2x_QNm5JE-f9nrDldhkgrilYzp4CFeYlznQ",
            "recipeUID": "37f0b96a9e8f2be46c5b70a2e58425b17c6f0a89",
            "capiArticleId": "lifeandstyle/2017/dec/08/christmas-cocktail-recipes-hawksmoor-hot-doddy-shaky-pete-gin-sherry",
            "sponsorshipCount": 0
        }
    ]
}
```
- Remove a character from one of the IDs and request again. You should see that the corresponding index entry disappears from the response and the `resolved` count drops.  The intention is that a client can see that one more more IDs is not valid by comparing the counts and if they are not equal should compare the list of returned `recipeUID` entries to the requested IDs if it needs to find the invalid ones

Fronts tool is able to translate UUIDs to checksums more easily

### DIFF
--- a/lambda/rest-endpoints/src/helpers.test.ts
+++ b/lambda/rest-endpoints/src/helpers.test.ts
@@ -1,6 +1,7 @@
+import {ProvisionedThroughputExceededException} from "@aws-sdk/client-dynamodb";
 import {v4 as uuid} from "uuid";
-import {getBodyContentAsJson, recursivelyGetIdList, validateDateParam} from "./helpers";
 import {multipleRecipesByUid} from "@recipes-api/lib/recipes-data";
+import {getBodyContentAsJson, recursivelyGetIdList, validateDateParam} from "./helpers";
 
 jest.mock("./config");
 jest.mock("@recipes-api/lib/recipes-data");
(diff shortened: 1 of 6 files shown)


lambda/rest-endpoints/src/helpers.test.ts
```diff
@@ -41,4 +47,62 @@ describe("app.validateDateParam", ()=>{
...
+    const results = await recursivelyGetIdList(uidList, []);
+    expect(results.length).toEqual(120);
+    expect(results.map(e=>e.recipeUID)).toEqual(uidList);
+    //we expect there to have been 3 parallel batches
+    expect((multipleRecipesByUid as jest.Mock).mock.calls.length).toEqual(3);
+  });
+
+  it("should retry if ProvisionedThroughputExceeded is caught", async ()=>{
+    (multipleRecipesByUid as jest.Mock).mockImplementation((idList:string[])=>{
+      const shouldCrash = Math.random();
```
JustinPinner, line 82: Is this `Math.random`isation reliable enough?

JustinPinner:
This reads well to me. Other than the question about using math.random to force a failure for the retry test (unless using the batch of 120 is _guaranteed_ to trigger it at least once) I'd be happy to approve it.

lambda/rest-endpoints/src/helpers.test.ts
fredex42: Well, the idea is that we crash ~5% of the time - on 120 resolves that should average 6 times so we'd have to be _really_ unlucky not to trigger at least one.  The test won't actually fail in this scenario; it's just extremely unlikely to happen

fredex42:
@JustinPinner Great thanks a lot. The block of code in question is called 120 times and should fail about 5% of the time which gives us a good enough window, the assertions are written loosely enough that the exact number of failures won't matter. But we'd be massively unlucky not to get at least one.  If it were to become a problem we could easily increase either the probability or number of calls

JustinPinner (approved):
👍

PR #65, opened 2024-09-17
Ensure eslint configuration integrates prettier properly

codingWithRach:
## What does this change?

Amends eslint configuration in an attempt to integrate Prettier properly, so that linting is consistent for all developers:
- change `extends` to include `"prettier"`
- adds `settings` section to specify `import/resolver` settings
- ignores `jest.config.ts` when linting

Also see changes described in PR #66, which was merged into this PR

As a result of improved linting, it was also necessary to fix 412 linting errors!

## How to test

`npm run lint` - no linting errors shown
`npm run test` - tests pass
`npm run build` - no errors

I guess this should also be deployed to CODE?  I welcome advice on how best to test it once deployed

> Yes I'll deploy to CODE, publish a recipe, and check the GET endpoints. POST endpoints are pretty much deprecated now along with Mise-En-Place so don't need to worry about those so much

## How can we measure success?

No more PRs where it's difficult to see the actual changes because of all the linting changes

## Have we considered potential risks?

Greater consistency is surely a good thing!

PR #66, opened 2024-09-17
Run prettier directly instead of via eslint

emdash-ie:
## What does this change?

While reviewing https://github.com/guardian/recipes-backend/pull/65, I started reading the prettier docs and found that they [recommend running prettier directly rather than via eslint](https://prettier.io/docs/en/integrating-with-linters.html). I started making some changes to suggest, but they turned out to be significant. To make them easier to review/compare, I’ve split them off into this PR.

The individual commit messages contain more info, which I’ve reproduced here:

### 1d04774 Remove prettier changes from eslintrc.jsons

Prettier’s docs [recommend running it directly rather than running it via eslint](https://prettier.io/docs/en/integrating-with-linters.html). To try it out, I’m first removing the existing eslint configuration.

### 524e92c Remove unnecessary .prettierrc

Since we’re using a prettier config specified in package.json, I believe this file isn’t actually used, so I’ve removed it. (The guardian prettier config also specifies the only rule in this file anyway.)

### 5ead7d1 Follow eslint-config-prettier setup instructions

Prettier’s recommended way of using it with eslint is to use [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) to disable any eslint rules that would conflict with prettier.

This commit sets up the eslint config as instructed in the readme on that repo.

### ~e837cda Update eslint-config-prettier~

~When attempting to install this dependency, I actually updated it, because it was already installed.~

### a80fc42 Add scripts to run prettier

I copied these [from dotcom-rendering](https://github.com/guardian/dotcom-rendering/blob/f9f1c28066c048a1b4c281514d69382f220d9a19/package.json#L19-L20) – I figure it’s probably good to be consistent with how they do things.

In particular, I’m adding these because I can’t get nx to run prettier with the same config as when I run it directly (I don’t even know if it’s definitely running prettier!): running `npm run format:check` with these changes warns that 102 files need formatting, whereas running `npm run nx format:check` only says two files need formatting (package-lock.json and package.json).

### a27b440 Run prettier check in CI

To ensure the files are formatted with prettier, this commit runs the format:check script in the github actions build.

### d060444 Format everything with prettier

To make this commit, I ran `npm run format:write`.

## How to test

- Test your editor integration: does it run prettier automatically when you make changes?
- Make some spacing edits to a file
  - You may need to disable your editor integration for this, or use nano or etc
- Confirm that `npm run format:check` reports those files need formatting
- Confirm that `npm run format:write` removes your spacing changes
- Push a commit with an incorrect formatting change to a PR, confirm that CI fails


codingWithRach:
That all seems like a very sensible approach!

codingWithRach:
Although there are loads of conflicts, probably caused by things that I've been working on in my original PR in the meantime (sorry!)

emdash-ie:
> Although there are loads of conflicts, probably caused by things that I've been working on in my original PR in the meantime (sorry!)

It’s no problem! I can remove the formatting commit, then rebase/cherry-pick if we want to use these changes.

PR #71, opened 2024-09-18
Improve the wording of the publication message

fredex42:
## What does this change?

The publication message we generate here is shown to users in the Fronts tool once publication is complete:

Screenshot 2024-09-18 at 18 19 18

At the moment it's fairly generic and does not tell use user what to expect.

This PR updates the message to tell them:
- if the issue date is today, that it should be live immediately
- if the issue date is in the future, which day it will go live
- if the issue date is in the past, that it can be seen in the fronts preview tool but will never actually be made live

## How to test

- Deploy to CODE
- Open the Northern Hemisphere / 18th Sep 2024 front (or generate your own with the new script 😀 )
- Publish it. See the message
- If you're particularly keen, generate some future ones to see that message too

Slightly nicer UX

### DIFF
--- a/lambda/facia-responder/src/util.test.ts
+++ b/lambda/facia-responder/src/util.test.ts
@@ -1,0 +1,16 @@
+import format from "date-fns/format";
+import {generatePublicationMessage} from "./util";
+
+describe("generatePublicationMessage", ()=>{
+  it("should generate a different message if the issueDate is today", ()=>{
+    const nowTime = new Date();
+    const formattedDateToday = format(nowTime, "yyyy-MM-dd");
+    const result = generatePublicationMessage(formattedDateToday);
+    expect(result).toEqual("This issue has been published and should be live in the app imminently");
+  });
+
+  it("should include the issue date if the issueDate is not today", ()=>{
+    const result = generatePublicationMessage("2023-02-10");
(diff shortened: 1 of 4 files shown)


lambda/facia-responder/src/util.test.ts
```diff
@@ -0,0 +1,21 @@
...
+  it("should generate a different message if the issueDate is today", ()=>{
+    const nowTime = new Date();
+    const formattedDateToday = format(nowTime, "yyyy-MM-dd");
+    const result = generatePublicationMessage(formattedDateToday);
+    expect(result).toEqual("This issue has been published and should be live in the app imminently");
+  });
+
+  it("should include the issue date if the issueDate is not today", ()=>{
+    const result = generatePublicationMessage("2023-02-10", new Date(2023, 1, 1));
+    expect(result).toContain("Fri, 6th Feb 2023")
```
JustinPinner, line 14: Not sure I understand why we'd expect Fri 6th Feb here when neither of the dates supplied are that date?

JustinPinner (approved):
One question otherwise looking good 👍

lambda/facia-responder/src/util.test.ts
fredex42: Thanks for catching this, the format string was wrong! Fixed now.

PR #72, opened 2024-09-27
Update Readme to reflect correct argument for recipeId

Divs-B:
## What does this change?

Their was a requirement came up by iOS devs to reindex a recipe. @paulmr and I tried running locally the script as mentioned in Readme and provided recipe id which ended up in an. error `TypeError [ERR_PARSE_ARGS_UNKNOWN_OPTION]: Unknown option '--recipe-uid'`

We found in the script that actually argument is `recipeUid` and not `recipe-uid` and hence changed it to make it work.

## How to test

Command should run without any error on recipe id.

re index should run

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,9 @@ You need to have a large number of environment variables set for the tools to wo
 
 1. Get CAPI credentials from Janus. You'll need the AWS commandline, and `jq` utility, installed; but you should have these already.
 2. Run `STAGE=CODE ./get-local-config.sh` to generate the file `environ-CODE`. Substitute PROD for CODE to get setup for the Production environment (but be careful!)
-   - I'd recommend deleting the file once you're done with it.
+
+- I'd recommend deleting the file once you're done with it.
+
 3. It goes without saying that these environ files should NEVER be committed to the repo. They are gitignored, but still... don't do it.
 4. Once you have the file, run `source environ-CODE` to set up the environment variables you need to target the CODE environment.
 5. You can now run the tools outlined below.
@@ -27,7 +29,7 @@ For a tighter feedback loop, run individual projects or files, and watch, with `
 ## How do I re-index content from CAPI out to Feast?
 
 1. Set up for local operations, as above
-2. Run `npm run commandline-reindex -- [--composer-id 1234567] [--capi-uri path/to/article/in/capi] [--recipe-uid 0551534c8d93e8da7bb70553b10fa0d0f62534a3]`
+2. Run `npm run commandline-reindex -- [--composerId 1234567] [--capiUri path/to/article/in/capi] [--recipeUid 0551534c8d93e8da7bb70553b10fa0d0f62534a3]`
 
 Note that Content API publishes _articles_, wheras we publish _recipes_. There may well be more than one recipe in an article. This command will therefore
 re-publish _every_ recipe from the given article; it may well change the SHA value (mutable ID) of some of the recipes. The index will be updated to respect this;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,26 +27,28 @@ For a tighter feedback loop, run individual projects or files, and watch, with `
 ## How do I re-index content from CAPI out to Feast?
 
 1. Set up for local operations, as above
-2. Run `npm run commandline-reindex -- [--composer-id 1234567] [--capi-uri path/to/article/in/capi] [--recipe-uid 0551534c8d93e8da7bb70553b10fa0d0f62534a3]`
+2. Run `npm run commandline-reindex -- [--composer-id 1234567] [--capi-uri path/to/article/in/capi] [--recipeUid 0551534c8d93e8da7bb70553b10fa0d0f62534a3]`
```
Divs-B, line 30: This one line change is actually for the PR others are auto linting i think.

JustinPinner: If I'm reading [this](https://github.com/guardian/recipes-backend/blob/07dae096fe7413f729bd5804f1b223c042843083/lambda/recipes-responder/src/commandline-reindex.ts#L86) correctly, all of these parameter values are camel case aren't they?

Divs-B: oh actually yes, others are also camel case. since I haven't ran any these arguments hence skipped checking them at the first place. Do you think we can amend these as well in this PR? 
Or, I hope it doesn't mean that letter should be camel case after single hyphen? like `--composer-id` should be passed as `composerId` and I have not checked it?

JustinPinner: Honestly I haven't dug into the details. Maybe `snake-case` params are automagically converted into `camelCase` somewhere and so it doesn't matter. The important thing is that the README is an accurate reflection of what we need to do. It just struck me as odd that we'd be illustrating a mixture of techniques as it means users will potentially have to refer to the README every time they want to use the script to remind themselves which param should be in which format - if it actually makes a difference, which I assume it does or there would be no need to make this change at all.

fredex42: OK, so the news is not good 😞 

Running this test script proves that there is no magic, and all the arge in this case are in fact camelCase (no idea how I missed that! sorry!)

```javascript
#!/usr/bin/env node

const { parseArgs } = require("node:util");

const {values} = parseArgs({
    options: {
        help: {
            type: 'boolean',
            short: 'h'
        },
        testOne: {
            type: 'string',
            short: 't'
        },
        testTwo: {
            type: 'string',
        }
    }
});

console.log(values);
```

```
$ node testargs.js --help --testOne Hello --testTwo 'everybody here'
[Object: null prototype] {
  help: true,
  testOne: 'Hello',
  testTwo: 'everybody here'
}
```

```
node testargs.js --help --test-one Hello --testTwo 'everybody here'
node:internal/util/parse_args/parse_args:98
    throw new ERR_PARSE_ARGS_UNKNOWN_OPTION(
    ^

TypeError [ERR_PARSE_ARGS_UNKNOWN_OPTION]: Unknown option '--test-one'
    at new NodeError (node:internal/errors:405:5)
    at checkOptionUsage (node:internal/util/parse_args/parse_args:98:11)
    at node:internal/util/parse_args/parse_args:360:9
    at Array.forEach (<anonymous>)
    at parseArgs (node:internal/util/parse_args/parse_args:357:3)
    at Object.<anonymous> (/Users/andy_gallagher/testargs.js:5:18)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12) {
  code: 'ERR_PARSE_ARGS_UNKNOWN_OPTION'
}
```

Behaviour is the same on Node v16.20.2, v18.18.2 and v22.6.0.

@Divs-B please could you update the other arguments in the README? Thanks a lot!

Divs-B: Thats a good finding to confirm on this, thankyou 👍 . 
I will update Readme for other `args` too.

PR #73, opened 2024-09-30
Test CI formatting check

emdash-ie:
This PR makes a number of weird formatting changes, to test whether the CI is checking formatting.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,10 +2,10 @@
 	"root": true,
 	"ignorePatterns": ["**/*.config.ts"],
 	"plugins": ["@nx"],
-	"extends": ["prettier"],
 	"overrides": [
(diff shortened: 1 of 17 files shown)


emdash-ie:
Added two more commits, showing the CI check is not complaining about odd spacing in the middle of the line (including a tab!), nor about trailing space at the ends of lines.

emdash-ie:
As a test, I’ve cherry-picked the commits from https://github.com/guardian/recipes-backend/pull/65 that originally configured eslint to run prettier. The build failures show that this successfully runs prettier with our config.

emdash-ie:
As a second test, I reverted the eslint config and restored the check that runs prettier directly – that also fails the build, correctly detecting the formatting errors.

PR #74, opened 2024-10-03
fix: Make re-indexing script more fault tolerant

fredex42:
## What does this change?

While re-indexing I came across a recipe in the index whose CAPI page did not resolve any more. This broke the entire re-index; so I made the error non-fatal.

It logs out an error then pauses for a couple of seconds so you can read it.

*Aside* - this is a rubbish way to re-index now we are properly in production. There is a card to do this properly.

## How to test

No need, I tested it already while making the PROD reindex work.

Able to re-index properly

n/a at present. We should consider how to handle these cases better when we implement the improved re-indexing procedure


Divs-B (approved):
Looks good 👍

PR #75, opened 2024-10-15
Fix eslint prettier integration

emdash-ie:
## What does this change?

When we merged https://github.com/guardian/recipes-backend/pull/65, we accidentally lost the necessary integration to run prettier as part of our eslint check! This PR fixes this by cherry picking two commits from that PR which re-enable the formatting check in CI.

## How to test

- check out the branch, add a commit with some weird formatting
- run `npm run lint` and verify that the check fails

I have also tested the CI check in https://github.com/guardian/recipes-backend/pull/73.

- our new PRs won’t include formatting changes
  - because the strict check in CI will prevent us introducing non-standard formatting
- we won’t have to think about indentation/formatting when writing code
  - because we can enable prettier in our IDEs, which will automatically format it for us


codingWithRach:
I ran `npm run lint` without making any changes to the formatting and it gave a load of errors.  So maybe some linting corrections were also lost along the way?

emdash-ie:
> I ran `npm run lint` without making any changes to the formatting and it gave a load of errors. So maybe some linting corrections were also lost along the way?

Yes, I’ve just noticed this as well :( Working on formatting the code now, and will see if I can find commits that fixed the other lint errors next.

codingWithRach:
This now works with `npm run lint`.  However, a slightly quirky thing is that when I run `npx nx format:check` it highlights issues in `package.json` and `package-lock.json`.  This almost definitely doesn't matter at all!

And there's an error with `npm run test`...

Sorry I didn't get all this sorted before I went on holiday!

emdash-ie:
> And there's an error with `npm run test`...

It’s weird, I don’t understand why these have broken?

emdash-ie:
> However, a slightly quirky thing is that when I run `npx nx format:check` it highlights issues in `package.json` and `package-lock.json`. This almost definitely doesn't matter at all!

We can probably fix those by running prettier – I’m running out of patience for getting nx/eslint to lint the files that matter to us.

codingWithRach:
> > And there's an error with `npm run test`...
> 
> It’s weird, I don’t understand why these have broken?

Nor I!  This isn't a file that needed to be fixed in #65 .  Hmmnn...

emdash-ie:
> > > And there's an error with `npm run test`...
> > 
> > 
> > It’s weird, I don’t understand why these have broken?
> 
> Nor I! This isn't a file that needed to be fixed in #65 . Hmmnn...

Ah wait, I was looking at the wrong bits of the output: having all the logs in the output even for passed tests is quite confusing. I should be able to make some change to fix it, will take a look.

codingWithRach (approved):
Looking good!

PR #77, opened 2024-11-18
Update CDK and dependencies

fredex42:
## What does this change?

Bumps CDK version and associated libraries via `npm audit fix`

## How to test

- Build and deploy to CODE
- Author a recipe out to CODE and check it's in the backend. If it works, all good!


emdash-ie (approved):

cdk/lib/__snapshots__/recipes-backend.test.ts.snap
```diff
@@ -113,26 +113,23 @@ import urllib.request
...
 def handler(event: dict, context):
   response_status = "SUCCESS"
   error_message = ""
   try:
     props = event["ResourceProperties"]
-    bucket = props["BucketName"]
     notification_configuration = props["NotificationConfiguration"]
-    request_type = event["RequestType"]
     managed = props.get('Managed', 'true').lower() == 'true'
+    skipDestinationValidation = props.get('SkipDestinationValidation', 'false').lower() == 'true'
```
JustinPinner, line 125: Is this a change to keep something imposed by CDK happy? Or perhaps it's not important because this is just for the test?

cdk/lib/__snapshots__/recipes-backend.test.ts.snap
```diff
@@ -113,26 +113,23 @@ import urllib.request
...
     stack_id = event['StackId']
-
+    old = event.get("OldResourceProperties", {}).get("NotificationConfiguration", {})
```
JustinPinner, line 127: How long do we have to support old messages for? Assuming I'm reading the use of `old` correctly below. Again, maybe not that important if it's just testing it still works either way.

cdk/lib/__snapshots__/recipes-backend.test.ts.snap
```diff
@@ -205,7 +201,20 @@ def submit_response(event: dict, context, response_status: str, error_message: s
...
+  for key, value in json_obj.items():
+      if isinstance(value, dict):
+          json_obj[key] = sort_filter_rules(value)
+      elif isinstance(value, list):
+          json_obj[key] = [sort_filter_rules(item) for item in value]
+  if "Filter" in json_obj and "Key" in json_obj["Filter"] and "FilterRules" in json_obj["Filter"]["Key"]:
+      filter_rules = json_obj["Filter"]["Key"]["FilterRules"]
+      sorted_filter_rules = sorted(filter_rules, key=lambda x: x["Name"])
+      json_obj["Filter"]["Key"]["FilterRules"] = sorted_filter_rules
+  return json_obj",
```
JustinPinner, line 217: Oof, this is a toughie. I'm not sure I can tell what this is doing, other than seemingly recursively calling itself until it ends up with a sorted list of `FilterRules` if applicable. Will have to try to figure out what that means.

cdk/lib/__snapshots__/recipes-backend.test.ts.snap
fredex42: Yes, it's in the snapshot; it's an artefact of one of the library bumps. Bit rubbish to get lumps of Python code in there but it's out of our control 🤷

cdk/lib/__snapshots__/recipes-backend.test.ts.snap
fredex42: likewise

PR #78, opened 2024-11-18
Supply a cache of information about contributor profiles to the app

fredex42:
## What does this change?

- Adds a new static endpoint, `v2/contributors.json`, which contains information about valid "chef" contributors for the app to use.
- This is built hourly by using the recipe search backend to determine all profile tags for which there is at least one recipe, then efficiently resolving them through CAPI and outputting the result as JSON into S3

This is a request from the Feast team in order to move the app to needing less initial data to be downloaded.  At present, they do this resolution client-side and need _every_ recipe downloaded in order to build the list of profile tags to resolve.  By doing this work for them up-front and giving them the data in json format, we can simplify matters considerably so they can simply "download chefs" and have done with it, lazy-loading recipes as required.

## How to test

- Deploy to CODE
- Find the `PublishRecipeContributors-CODE` lambda function in AWS console
- Use the 'Test' panel to manually run it
- Watch the logs - no errors reported. It should report that it found about 15 contributors from the recipe search and resolved 10 of them in CAPI (this is CODE, CAPI does not have anywhere near all the contributors in the CODE environment).
- Use curl, Postman, browser etc. to check https://recipes.code.dev-guardianapis.com/v2/contributors.json.
  - You should see the right amount of data there - including id, name, bio, image, etc.
  - The `Cache-Control` header should indicate a very long `max-age` (`max-age=31557600; stale-while-revalidate=60; stale-if-error=300`)
  - You should probably see the `X-Cache` header to be `HIT` - the cache is only flushed if a real update to the file is made
  - Repeating the request with a request header of `If-Modified-Since` and a recent date (use the same time format as the outgoing `Date` header) should show a response of `304 Not Modified` with no body content

Removing a blocker from the apps lazy-loading data

This does publicly expose information about contributors. However, this information is already publicly exposed through CAPI and the website so there is not actually any change.

### DIFF
--- a/cdk/lib/recipes-backend.ts
+++ b/cdk/lib/recipes-backend.ts
@@ -140,6 +140,11 @@ export class RecipesBackend extends GuStack {
 				? 'recipes.code.dev-guardianapis.com'
 				: 'recipes.guardianapis.com';
 
+		const capiUrlBase =
+			this.stage === 'CODE'
+				? 'content.code.dev-guardianapis.com'
+				: 'content.guardianapis.com';
+
 		const eventBus = EventBus.fromEventBusName(
 			this,
 			'CrierEventBus',
@@ -334,6 +339,45 @@ export class RecipesBackend extends GuStack {
 			new LambdaDestination(publishTodaysCurationLambda),
 			{ suffix: 'curation.json' },
 		);
+
+		new GuScheduledLambda(this, 'PublishContributors', {
+			app: 'recipes-publish-contributor-information',
+			architecture: Architecture.ARM_64,
+			fileName: 'profile-cache-rebuild.zip',
+			functionName: `PublishRecipeContributors-${this.stage}`,
+			handler: 'main.handler',
+			initialPolicy: [
+				new PolicyStatement({
+					effect: Effect.DENY,
+					actions: ['*'],
+					resources: [serving.staticBucket.bucketArn + '/content/*'],
+				}),
+				new PolicyStatement({
+					effect: Effect.ALLOW,
+					actions: ['s3:PutObject', 's3:GetObject'],
+					resources: [serving.staticBucket.bucketArn + '/*'],
+				}),
+			],
+			memorySize: 256,
+			monitoringConfiguration: {
+				noMonitoring: true, //TBD
+			},
+			rules: [
+				{
+					schedule: Schedule.cron({ minute: '16' }),
+					description:
+						'Update cache of contributor information for Feast at 16 minutes past every hour',
+				},
+			],
+			runtime: Runtime.NODEJS_20_X,
+			timeout: Duration.seconds(30),
+			environment: {
+				STATIC_BUCKET: serving.staticBucket.bucketName,
+				FASTLY_API_KEY: fastlyKeyParam.valueAsString,
+				BASE_URL: contentUrlBase,
+				CAPI_BASE_URL: capiUrlBase,
+			},
+		});
 
 		new DynamicFronts(this, 'DynamicFronts', {
 			destBucket: serving.staticBucket,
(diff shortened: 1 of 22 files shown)


lib/recipes-data/src/lib/s3.ts
```diff
@@ -160,3 +160,22 @@ export async function writeIndexData(indexData: RecipeIndex, Key: string) {
...
+export async function writeChefData(chefData: ChefInfoFile, Key: string) {
+	console.log('Marshalling data...');
+	const formattedData = JSON.stringify(chefData);
+
+	console.log(`Done. Writing to s3://${Bucket}/${Key}...`);
+	const req = new PutObjectCommand({
+		Bucket,
+		Key,
+		Body: formattedData,
+		ContentType: 'application.json',
```
michaelclapham, line 173: I think this needs to be application/json instead of application.json

fredex42: doh! thanks for the typo spot 😁

cdk/lib/recipes-backend.ts
```diff
@@ -335,6 +340,45 @@ export class RecipesBackend extends GuStack {
...
+			functionName: `PublishRecipeContributors-${this.stage}`,
+			handler: 'main.handler',
+			initialPolicy: [
+				new PolicyStatement({
+					effect: Effect.DENY,
+					actions: ['*'],
+					resources: [serving.staticBucket.bucketArn + '/content/*'],
+				}),
+				new PolicyStatement({
+					effect: Effect.ALLOW,
```
Divs-B, line 356: I think we don't need to provide ALLOW property, recently I came across with the situation to confirm where I was able to get S3 work done without this.  
it is auto default to ALLOW i think, so just fyi and not to change here anything.

cdk/lib/recipes-backend.ts
```diff
@@ -335,6 +340,45 @@ export class RecipesBackend extends GuStack {
...
+			],
+			memorySize: 256,
+			monitoringConfiguration: {
+				noMonitoring: true, //TBD
+			},
+			rules: [
+				{
+					schedule: Schedule.cron({ minute: '16' }),
+					description:
+						'Update cache of contributor information for Feast at 16 minutes past every hour',
```
Divs-B, line 369: ah interesting very specific time value 😁
i know, this is to get it done in calm time to avoid other process in between 👍

lib/capi/src/lib/capitags.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * We need to have <2048 chars in the URI or one of the backends objects.
+ * So, we try to bundle all of the tag IDs into one lookup; if that fails, we split in half and try two; etc.
+ * @param tagIdList the list of tag IDs to resolve
+ * @param capiBaseUrl base URL for CAPI
+ * @param capiKey API key for CAPI
+ * @param tagIdListTail remaining tag IDs to process next iteration. Internal use only.
+ * @param prevUrlList URLs already determined. Internal use only.
+ */
+export function buildUriList(
```
Divs-B, line 15: nice 👍 
i was curious to check and it suggested on that, that even if `tagIdListTail` 's length is higher than URL_MAX it will still divide in small chunks using recursive way

Divs-B (approved):
Code makes sense to me, 
Tested by hitting at url endpoint and i can see chef data with model this PR owns.
worth to parse thrift to our chef model to facilitate apps side in so many ways 🌟

lib/capi/src/lib/capitags.ts
fredex42: Exactly, it splits in half and recurses if the list is too long

cdk/lib/recipes-backend.ts
fredex42: Yeah try to avoid _everything_ hitting on the hour!

PR #79, opened 2024-11-20
Add step function for recipe reindex

jonathonherbert:
## Accessibility


cdk/lib/recipes-backend.ts
```diff
@@ -230,6 +241,11 @@ export class RecipesBackend extends GuStack {
 			dataStore: store,
 		});
 
+		new RecipesReindex(this, 'RecipeReindex', {
+			contentUrlBase,
+			reindexBatchSize: reindexBatchSizeParam.valueAsString,
```
JustinPinner, line 246: Do we need any type and/or availability guards around this?

JustinPinner:
To summarise what I _think_ I read here: The _entire_ recipe index is snapshotted to an S3 bucket as a lump of JSON. The reindex lambda pulls `<configurable_value>` records from the `Recipes` object in said JSON, extracts their IDs and then... This is where I think I'm missing something as I was expecting to see an interaction with the aforementioned queue (SQS?)

jonathonherbert:
> To summarise what I _think_ I read here: The _entire_ recipe index is snapshotted to an S3 bucket as a lump of JSON. The reindex lambda pulls `<configurable_value>` records from the `Recipes` object in said JSON, extracts their IDs and then... This is where I think I'm missing something as I was expecting to see an interaction with the aforementioned queue (SQS?)

That's right — EventBridge integration is the next step! I mention it briefly in the PR description, but it's a long one 😀

cdk/lib/recipes-backend.ts
jonathonherbert: I don't think we can guard against this being a string if we're pulling it from param store — there's no number type there. A few things we can do:

1. Keep this as a stack param, where we can specify a number type. + catch failure at input time (preferable), - require a manual stack update to update (small risk of unintended changes)
2. Better failure mode when this isn't a `number`. + a bit clearer what is happening in logs, - failure on run is much less transparent

Because this is an SSM param, and it is likely to change infrequently, I'm leaning towards 2 — but WDYT?

JustinPinner: 2 sounds good to me, too. Protecting ourselves from it not being a number, or being missing entirely when we try to access it is good 👍

fredex42: This code does only run at build time (it's in the CDK) so a failure won't affect the running app

cdk/lib/recipes-reindex.ts
```diff
@@ -0,0 +1,182 @@
...
+		scope: GuStack,
+		id: string,
+		{ contentUrlBase, reindexBatchSize }: RecipesReindexProps,
+	) {
+		super(scope, id);
+
+		const appBase = 'recipes-reindex';
+		const lambdaFileName = `${appBase}.zip`;
+
+		const snapshotBucket = new Bucket(this, 'staticServing', {
```
fredex42, line 37: little niggle - can we get a better name than `staticServing`? Something like `reindexSnapshots`?

cdk/lib/recipes-reindex.ts
```diff
@@ -0,0 +1,182 @@
...
+
+		const stateMachine = new StateMachine(this, 'ReindexingStateMachine', {
+			definition,
+			timeout: Duration.minutes(15),
+		});
+
+		stateMachine.addToRolePolicy(
+			new PolicyStatement({
+				effect: Effect.ALLOW,
+				resources: ['*'],
```
fredex42, line 177: I don't suppose we can narrow this down at all?

lambda/recipes-reindex/README.md
```diff
@@ -0,0 +1,7 @@
+# lambda/recipes-reindex
+
+This library was generated with [Nx](https://nx.dev).
+
+## Running unit tests
+
+Run `nx test lambda/recipes-reindex` to execute the unit tests via [Jest](https://jestjs.io).
```
fredex42, line 7: Could be removed?

lambda/recipes-reindex/src/snapshotRecipeReindex/index.ts
```diff
@@ -0,0 +1,53 @@
...
+	const { ContentUrlBase, RecipeIndexSnapshotBucket } = getConfig();
+	const { executionId } = state;
+
+	const recipeIndexUrl = `https://${ContentUrlBase}/${INDEX_JSON}`;
+	console.log(`Fetching recipe index from ${recipeIndexUrl}`);
+	const recipeIndexSnapshotResponse = await fetch(recipeIndexUrl);
+
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- absent Zod types for the index, casting is necessary
+	const recipeIndexSnapshotJson: RecipeIndex =
+		await recipeIndexSnapshotResponse.json();
```
fredex42, line 27: Another option is to use `retrieveIndexData` from `lib/recipes-data` which is already typed (retrieves directly from dynamo)

lambda/recipes-reindex/src/writeBatchToReindexQueue/config.ts
```diff
@@ -0,0 +1,7 @@
+import { mandatoryParameter } from 'lib/recipes-data/src/lib/parameters';
+import { RecipeIndexSnapshotBucket } from '../sharedConfig';
+
+export const getConfig = () => ({
+	RecipeIndexSnapshotBucket,
+	ReindexBatchSize: parseInt(mandatoryParameter('REINDEX_BATCH_SIZE')),
```
fredex42, line 6: Do we guard against this returning e.g. `NaN`?

tsconfig.base.json
```diff
@@ -17,6 +17,9 @@
 		"baseUrl": ".",
 		"paths": {
 			"@recipes-api/cwmetrics": ["lib/cwmetrics/src/index.ts"],
+			"@recipes-api/lambda/recipes-reindex": [
+				"lambda/recipes-reindex/src/index.ts"
+			],
```
fredex42, line 22: Not sure that i understand this - is the package intended as a library or an executable?

fredex42:
Another little thing - might be nice to set a "friendly name" on the step function state machine? `
RecipeReindexReindexingStateMachine721AD70A-UpylgSoIFutc` is a little ugly...

cdk/lib/recipes-reindex.ts
jonathonherbert: Such a good question. I tried, good lord I tried, to scope this to the SM in question, and could not avoid CDK complaining about a circular dependency.

Having said that, I've just re-searched our estate and it looks like it might be possible to do this by reconstructing the arn manually — bear with ...

lambda/recipes-reindex/src/writeBatchToReindexQueue/config.ts
jonathonherbert: We don't do this [anyplace else in config](https://github.com/guardian/recipes-backend/blob/11aba448643d99974fee671868f4b664f2b54798/lib/recipes-data/src/lib/config.ts#L21) AFAICS, but we should probably fail fast in these cases! Another PR to address this more thoroughly across all apps? cc. @JustinPinner, as it relates to the comment above.

tsconfig.base.json
jonathonherbert: Mistakenly created as a lib first, and missed this on cleanup, thanks for spotting. dc4ceac0892184524353edc87c216f0b3545a4c4

cdk/lib/recipes-reindex.ts
jonathonherbert: Ironically your other comment https://github.com/guardian/recipes-backend/pull/79#issuecomment-2515093595 is the solution to this — by manually defining a name and manually constructing the result ARN we can avoid a forward reference 👍  dd5ece910680dc7f337fdc1fac58f16e9b00a9ee

cdk/lib/recipes-reindex.ts
jonathonherbert: Ctrl-C snafu. aeee945ae5fe9728a77db534844b7eafbb733ac0 👍

fredex42 (commented):
removing lock as i know you've seen it 😁

lambda/recipes-reindex/src/snapshotRecipeReindex/index.ts
jonathonherbert: I gave this a bash, and it works well but requires a large refactor to ensure we don't pull in irrelevant fail-fast params. I'll ship this as-is, but it'd be nice to come up with a strategy for configuration that ensured there were no surprises when an import unexpectedly implicated environment parameters. I've a few ideas, as I'm sure you do too 😀

cdk/lib/recipes-backend.ts
jonathonherbert: I've gone with 1. for now, as I can't see a way to transform the string param from SSM into a number for the wait fn's `Duration.seconds` constructor (you can go the other way with `stringifyNumber`, but there's no number -> string equivalent AFAICS?) — we can revisit as needed. bc2d9a722e82f55f379430486c51095adeba7b27

fredex42: @jonathonherbert this is because `reindexBatchSizeParam.valueAsString` does _not_ actually return a string. Rather, it gives a "placeholder" which is rendered into the Cloudformation as the equivalent of:

```yaml
reindexBatchSize: !Ref reindexBatchSizeParam
```

This is an advantage because it's evaluated at deploy time not build time.

Therefore there is not way to transform it to a number, because Cloudformation can't support that.

fredex42 (approved):

cdk/lib/recipes-backend.ts
jonathonherbert: I know, it's a token: I suppose there's a way to move from numbers to strings because that is always possible, whereas moving in the other direction can fail 💥

PR #80, opened 2024-11-21
Deprecate MiseEnPlace POST endpoint

Divs-B:
## What does this change?

We don't need to hold `Misenplace` references to curate data because we have `Fronts-tools` fully functioning and working efficiently for the same.
reference of the task: https://trello.com/c/JdzlfTbD/3415-deprecate-mise-en-place-post-endpoints

## How to test

Mentioned in PR below how we tested, 
We tested POST and GET request to see if they are working as expected and it seems they are.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,10 @@ most of the actual logic lives in the library code imported into it
 
 A lambda function that rebuilds the index on-demand. This is incorporated from initial testing and will probably be removed.
 
+### lambda/rest-endpoints
+
+A lambda function initially used to POST the curation data for MEP to use it, now exists to support GET endpoint to get the most recent version of a given recipe ID, used by Fronts tool to resolve the recipe unique id to a checksum
+
 ### lib/capi
 
 Library functions to communicate with the Content Application Programmer's Interface. This is imported into the lambda code as `@recipes-api/lib/capi`.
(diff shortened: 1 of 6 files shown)


Divs-B:
#### Execute POST curation.json to any date:
`curl -X POST https://recipes.code.dev-guardianapis.com/northern/all-recipes/2024-10-16 -d@curation.json --header 'Content-Type: application/json' --header 'x-api-key: my-api-key' 
#### Returning response:
`<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>MethodNotAllowed</Code><Message>The specified method is not allowed against this resource.</Message><Method>POST</Method><ResourceType>OBJECT</ResourceType><RequestId>Z8CEXNG1JP9A3RYE</RequestId><HostId>J69E6Hd1b+/KoA41V3c6dZ2OGDzOmzb24eY7iSIqvn4YSNF8eMbjXc0BpXEM6FGOZ7Og05ERhfooMUgTzqnTopy+wJBWjeq22ZVaW+iB/zU=</HostId></Error>% `

fredex42:
Apologies @Divs-B , I know I said to remove the data models but in some unrelated work I ended up needing them as it's still the data format that the app uses.  I've reinstated them and renamed to `FeastAppCurationPayload`

fredex42 (approved):
😁

PR #83, opened 2024-12-05
Add a new transform to temporarily fix `celebration(s)Ids`

fredex42:
## What does this change?

- Duplicates the contents of `celebrationIds` to `celebrationsIds`, if the former array exists and has at least one member.
- This is to allow older versions of the app to correctly read the data, this can be removed once enough app users have updated
- Since this is on ingest, a re-index will be required

## How to test

- Transformation code is tested in CI

- Use the recipe search API on CODE to see how many recipes have celebrationIds on them
- Re-index recipes on CODE
- Re-index search on CODE
- Check some of the recipes - ones that had `celebrationIds` should now have both `celebrationIds` and `celebrationsIds` and have their checksum updated. Ones that did not have any `celebrationIds` should be unchanged

_I have already carried this out on CODE_

App is able to search for christmas recipes!

Since this involves re-indexing we need to be aware of forcing users to re-download a large delta.  We've checked the stats and it only looks like we'll be updating 600-700 recipes out of 5000 and this is considered acceptable.

### DIFF
--- a/lib/recipes-data/src/lib/extract-recipes.ts
+++ b/lib/recipes-data/src/lib/extract-recipes.ts
@@ -17,6 +17,7 @@ import {
 	handleFreeTextContribs,
 	replaceCanonicalArticle,
 	replaceImageUrlsWithFastly,
+	temporaryCelebrationIdsFix,
 } from './transform';
 import type {
 	RecipeTransformationFunction,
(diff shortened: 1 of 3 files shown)


JustinPinner (approved):
LGTM 👍

PR #85, opened 2024-12-10
Retrievable tag bugfix

fredex42:
## What does this change?

- Fixes a nasty bug where the "retrievable update" code (also used in the reindexing process) was not requesting tag data from CAPI resulting in a loss of sponsorship information
- Slight improvement to the reindex script code so we log failures at the end rather than interrupting partway through

## How to test

- Deploy to CODE
- Perform a reindex on e.g. da5442fb9ed648df95c5b5bd592bef3b
```bash
npm run commandline-reindex -- --recipeUid=da5442fb9ed648df95c5b5bd592bef3b
```
- Request https://recipes.code.dev-guardianapis.com/v2/index.json
- Request it again (the first one is usually served stale content while the CDN refreshes in the background)
- Find that recipe in the index, it should still have `"sponsorshipCount": 1`
- Request https://recipes.code.dev-guardianapis.com/content/da5442fb9ed648df95c5b5bd592bef3b
- This should have the sponsorship data that was there before

Don't accidentally lose data!

n/a

### DIFF
--- a/lambda/recipes-responder/src/commandline-reindex.ts
+++ b/lambda/recipes-responder/src/commandline-reindex.ts
@@ -166,6 +166,7 @@ async function main() {
 		else console.log(msg);
 		console.log('------------------------------------------------------\n');
 	}
+	const failedArticleIds: string[] = [];
 
 	if (all && !indexOnly) {
 		const index = await retrieveIndexData();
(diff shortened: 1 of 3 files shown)


jonathonherbert (approved):
The core change makes sense so approving to unblock.

PR #86, opened 2024-12-12
Refactor configuration getters

jonathonherbert:
## What does this change?

Refactors our config in recipes-backend to ensure that importing utilities does not accidentally pull in mandatory parameters. 

This is a side-quest as I attempted to include dynamoDB helpers into the reindex lambda. At the moment, importing these helpers also pulls in a host of exported vars that call getMandatoryParam and fail fast, but are not required in a reindex context.

The solution in this PR is that all mandatory configuration variables become getters that must be called.

Each lambda handler invokes these functions at its beginning, and passes values explicitly in named object fields. Tests provide the appropriate values explicitly.

This is

- boring 😀 
- slightly more verbose
- harder to get wrong (by naming arguments: having functions with large arities and many arguments of type `string` means argument order matters, and neither the compiler nor common sense can warn us when we are passing arguments in an incorrect order at the point of invocation)
- significantly less complicated to debug at runtime.

If we want smaller function arities, we could refactor our fns to receive objects (the JS equivalent of 'named parameters'.)

## How to test

- The tests should pass.
- Deploy to CODE, and exercise publication and curation paths. They should work as expected.

Testing on CODE, verifying that behaviour is correct and Fastly is purging for the following operations:

- [x] publish new recipe
- [x] update recipe
- [x] take down 
- [x] update curation
- [x] manual reindex

No surprises when importing configuration from libraries into lambda projects.

### DIFF
--- a/lib/recipes-data/src/lib/fastly.ts
+++ b/lib/recipes-data/src/lib/fastly.ts
@@ -35,12 +35,17 @@ function removeLeadingAndTrailingSlash(from: string): string {
  * @param apiKey Fastly API key to authenticate the request.  The function will throw if this is undefined or empty.
  * @param purgeType Whether to execute a soft or hard purge (default soft). See the docs on PurgeType for more information.
  */
-export async function sendFastlyPurgeRequest(
-	contentPath: string,
-	apiKey: string,
-	contentPrefix: string,
-	purgeType?: PurgeType,
-) {
+export async function sendFastlyPurgeRequest({
+	contentPath,
+	apiKey,
+	contentPrefix,
+	purgeType,
+}: {
+	contentPath: string;
+	apiKey: string;
+	contentPrefix: string;
+	purgeType?: PurgeType;
+}) {
 	if (!apiKey || apiKey == '') {
 		throw new Error('Cannot purge because Fastly API key is not set');
 	}
@@ -98,20 +103,26 @@ export async function sendFastlyPurgeRequest(
  * @param purgeType Whether to execute a soft or hard purge (default soft). See the docs on PurgeType for more information.
  * @param retryCount don't specify this, it's used internally.
  */
-export async function sendFastlyPurgeRequestWithRetries(
-	contentPath: string,
-	apiKey: string,
-	contentPrefix: string,
-	purgeType?: PurgeType,
-	retryCount?: number,
-): Promise<void> {
+export async function sendFastlyPurgeRequestWithRetries({
+	contentPath,
+	apiKey,
+	contentPrefix,
+	purgeType,
+	retryCount,
+}: {
+	contentPath: string;
+	apiKey: string;
+	contentPrefix: string;
+	purgeType?: PurgeType;
+	retryCount?: number;
+}): Promise<void> {
 	try {
-		return sendFastlyPurgeRequest(
+		return sendFastlyPurgeRequest({
 			contentPath,
 			apiKey,
 			contentPrefix,
 			purgeType,
-		);
+		});
 	} catch (err) {
 		if (err instanceof FastlyError) {
 			const nextRetry = retryCount ? retryCount + 1 : 1;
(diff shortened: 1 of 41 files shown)


jonathonherbert:
This PR is not yet tested end-to-end — although reviews are welcome, I won't ship it until I've given it a bit of a rinse in CODE to double-check I haven't fat fingered param orders.

jonathonherbert:
> fat-fingered param orders

no longer possible with named arguments: [18ac829](https://github.com/guardian/recipes-backend/pull/86/commits/18ac829f342d920ff4a9affb8e66384986a644d4)

lib/recipes-data/src/lib/fastly.ts
```diff
@@ -35,21 +35,24 @@ function removeLeadingAndTrailingSlash(from: string): string {
...
-	}
+export async function sendFastlyPurgeRequest({
+	contentPath,
+	apiKey,
+	contentPrefix,
+	purgeType,
+}: {
+	contentPath: string;
+	apiKey: string;
+	contentPrefix: string;
```
fredex42, line 46: should probably update the docs here too 😁

lib/recipes-data/src/lib/fastly.ts
```diff
@@ -100,14 +103,26 @@ export async function sendFastlyPurgeRequest(
...
+export async function sendFastlyPurgeRequestWithRetries({
+	contentPath,
+	apiKey,
+	contentPrefix,
+	purgeType,
+	retryCount,
+}: {
+	contentPath: string;
+	apiKey: string;
+	contentPrefix: string;
```
fredex42, line 115: again

lib/recipes-data/src/lib/parameters.ts
```diff
@@ -1,13 +1,31 @@
-export function mandatoryParameter(name: string): string {
-	if (process.env[name]) {
-		return process.env[name] as string;
-	} else {
-		if (process.env['CI']) {
-			return 'test';
+export function createGetMandatoryParameter<AliasType extends string>(
+	name: string,
+): () => AliasType {
```
fredex42, line 3: Question - what's the advantage of defining `AliasType` here, when the type of the return value is guaranteed to be a concrete `string` (either the result of `process.env[name]` or `'test'`)?

lib/recipes-data/src/lib/s3.ts
```diff
@@ -91,28 +108,41 @@ export async function publishRecipeContent(
...
+	purgeType?: PurgeType;
+	attempt?: number;
+}): Promise<void> {
 	const realAttempt = attempt ?? 1;
 
 	const Key = `content/${recipeSHA}`;
-	console.debug(`DEBUG: removeRecipeContent path is s3://${Bucket}/${Key}`);
+	console.debug(
+		`DEBUG: removeRecipeContent path is s3://${staticBucketName}/${Key}`,
+	);
```
fredex42, line 131: could probably ⚡ the debug line

fredex42 (approved):
couple of minor asks but generally a 👍  from me

lib/recipes-data/src/lib/parameters.ts
jonathonherbert: No advantage, this was me forgetting that in Typescript it's 🦆  all the way down: removed in 4185046f4c5b70dc0f8fdde462a26c754484cd7f

lib/recipes-data/src/lib/fastly.ts
jonathonherbert: 👍 4185046f4c5b70dc0f8fdde462a26c754484cd7f

lib/recipes-data/src/lib/s3.ts
jonathonherbert:
```suggestion
```

PR #87, opened 2024-12-19
Write article ids for reindex

jonathonherbert:
## What does this change?

Write article ids in our snapshot, rather than the recipe index; it's all we need to provide to the reindex queue.

## Why didn't we do this before?

Before this PR, the assumption was that we would reindex on a per-recipe basis.

However, we fetch recipes from our reindex source, CAPI, on a per-article basis.

It's inefficient, for an article that contains many recipes, to fetch that article many times to retrieve each recipe for a reindex. So we now reindex recipes on a per-article basis, to ensure we only fetch each article once for all the recipes it contains.

## How to test

- Observe that the code and test fixture have been refactored to pull data straight from dynamoDB, and write a list of article paths to S3 (rather than the entire `RecipeIndex` structure.
- Deploy to CODE, and exercise the step function. It should run successfully, and log healthy output:

Screenshot 2024-12-19 at 15 14 13

### DIFF
--- a/cdk/lib/__snapshots__/recipes-backend.test.ts.snap
+++ b/cdk/lib/__snapshots__/recipes-backend.test.ts.snap
@@ -2413,6 +2413,10 @@ def sort_filter_rules(json_obj):
           "Variables": {
             "APP": "recipes-reindex-snapshot-recipe-index",
             "CONTENT_URL_BASE": "recipes.guardianapis.com",
+            "INDEX_TABLE": {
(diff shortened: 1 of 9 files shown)


Divs-B (approved):
This is making sense to me. Also I'v verified the snapshot which generated after step function cycle ran this morning, the collated data is as expected in the snapshot.
Thanks for walk-through and clearing my doubts @jonathonherbert.

fredex42 (approved):
Looks good to me! Thanks a lot @jonathonherbert

PR #88, opened 2024-12-19
Write to event bus

jonathonherbert:
## What does this change?

Makes a few changes to the reindex code and infra to enable reindexing on a per-article basis:
- has the write events lambda ... write events to the event bus (which is doing much more than `crier` at this stage — perhaps due a rename)
- sets up a rule to pass these events to the `recipes-responder` lambda
- has the recipes responder discriminate between crier events and reindex events, and respond accordingly. I've tried to make the smallest possible change in the responder to exercise the same code path as retrievable updates, in this case renaming `handleContentUpdateRetrievable` -> `handleContentUpdateByCapiUrl`.

This should makes reindexing recipes by article possible. Happily (I think?), the recipes-responder will not emit writes when the recipe SHA hasn't changed — I've tested and recipes-responder does still emit events for recipe search to pick up, so reindexing should push all recipes through search, too.

## But it's so slow!

We're currently sending 10 articles at a time every 10 seconds from the batch writer lambda. Both these numbers can be tweaked by updating the stack params `reindexBatchSizeParam` and `reindexWaitTimeParam`. At the moment, it takes about half a second for recipes-responder to do everything it needs to do, from event receipt to final purge of index.json, so we could probably add a `0` to `reindexBatchSizeParam` without problems. But:

**Things to be aware of as we tweak these numbers:**
- the recipe-responder fn currently uses unreserved account concurrency. This means many responder fns can run in parallel. This is not desirable, because it could lead to race conditions as concurrent invocations compete to write their version of the final index to `index.json`. Perhaps we restrict the responder lambda to a single invocation at a time — or have a separate writer lambda for `index.json` that has this restriction — to prevent this condition. Which leads to another thought:
- there's no means of handling backpressure outside of AWS's own infrastructure at present. If our batch writer lambda is pushing records faster than downstream can handle them, we don't have visibility of that fact (other than carefully reading logs) — we cannot see whether messages are being dropped, and we cannot clear backlogs as we would be able to in e.g. SQS. Perhaps we can consider adding a queue in front of our responder lambda, which would solve both these problems.

Both of these are edge cases during normal operation, but in situations where we would like to recover data, or recover from a bad model change that is breaking Feast clients, they may become very important.

## Some questions

When reindexing, I saw some slightly odd behaviour in CODE. I think the reindex is working correctly, but it's worth checking that this is what we expect to see:

- Content that is available on open, but not on  Feast channel, does not appear in Feast. Is this intended?
- Some content in CODE has been taken down from Feast channel, but is still present in recipes-backend — e.g. lifeandstyle/2016/feb/21/ofm-classic-cookbooks-margaret-costa-four-seasons-cookery-book. Presume this is should never happen: we’d expect this content to be deleted in recipes-backend via Crier. Should reindex remove these items with a warning if it finds them?

## How to test

- The automated tests should pass.
- Run a reindex in CODE!
   - Go to step functions in CAPI account, and find [recipes-reindex-CODE-reindex-recipes](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/statemachines/view/arn:aws:states:eu-west-1:308506855511:stateMachine:recipes-reindex-CODE-reindex-recipes)
   - Start a new execution, adding `{ "dryRun": false }` to the payload (unless you'd like to test dryRun behaviour)
   - Watch the reindex working, perhaps by inspecting the [write lambda logs](https://logs.gutools.co.uk/s/content-platforms/app/r/s/1MBQj) and the [responder logs](https://logs.gutools.co.uk/s/content-platforms/app/r/s/n5wpJ) simultaneously.

## What's next

This PR implements reindexing, but doesn't provide the state necessary to let infra downstream of the step fn understand the identity of a reindex, whether a reindex is in progress, and how far we're through it. This PR was large enough that I thought it'd be best to implement that in another PR.

### DIFF
--- a/cdk/lib/recipes-backend.ts
+++ b/cdk/lib/recipes-backend.ts
@@ -229,6 +229,24 @@ export class RecipesBackend extends GuStack {
 			],
 		});
 
+		new Rule(this, 'ReindexConnection', {
+			eventBus,
+			description: `Connect recipe responder ${this.stage} to recipes-reindex`,
+			eventPattern: {
+				source: ['recipes-reindex'],
+				detail: {
+					channels: ['articles'],
+				},
+			},
+			targets: [
+				new aws_events_targets.LambdaFunction(updaterLambda, {
+					deadLetterQueue: responderDLQ,
+					maxEventAge: Duration.minutes(30),
+					retryAttempts: 5,
+				}),
+			],
+		});
+
 		new FaciaConnection(this, 'RecipesFacia', {
 			fastlyKeyParam,
 			serving,
@@ -253,6 +271,7 @@ export class RecipesBackend extends GuStack {
 			contentUrlBase,
 			reindexBatchSize: reindexBatchSizeParam.valueAsNumber,
 			reindexWaitTime: reindexWaitTimeParam.valueAsNumber,
+			eventBus,
 		});
 
 		const durationAlarm = new Alarm(this, 'DurationRuntimeAlarm', {
--- a/cdk/lib/recipes-backend.ts
+++ b/cdk/lib/recipes-backend.ts
@@ -234,9 +234,6 @@ export class RecipesBackend extends GuStack {
 			description: `Connect recipe responder ${this.stage} to recipes-reindex`,
 			eventPattern: {
 				source: ['recipes-reindex'],
-				detail: {
-					channels: ['articles'],
-				},
 			},
 			targets: [
 				new aws_events_targets.LambdaFunction(updaterLambda, {
--- a/cdk/lib/recipes-backend.ts
+++ b/cdk/lib/recipes-backend.ts
@@ -166,6 +166,10 @@ export class RecipesBackend extends GuStack {
 				TELEMETRY_XAR: telemetryXAR.valueAsString,
 				TELEMETRY_TOPIC: telemetryTopic.valueAsString,
 				OUTGOING_EVENT_BUS: eventBus.eventBusName,
+				CAPI_BASE_URL:
+					this.stage === 'PROD'
+						? 'https://content.guardianapis.com'
+						: 'https://content.code.dev-guardianapis.com',
 			},
 			initialPolicy: [
 				new PolicyStatement({
--- a/cdk/lib/recipes-reindex.ts
+++ b/cdk/lib/recipes-reindex.ts
@@ -127,6 +127,16 @@ export class RecipesReindex extends Construct {
 			},
 		);
 
+		// Necessary to reference `dryRun` later in the step fn.
+		const storeDryRun = new CustomState(scope, 'storeDryRun', {
+			stateJson: {
+				Type: 'Pass',
+				Assign: {
+					'dryRun.$': '$.dryRun',
+				},
+			},
+		});
+
 		const checkForOtherRunningReindexesTask = new CustomState(
 			scope,
 			'checkForOtherRunningTasks',
@@ -160,6 +170,10 @@ export class RecipesReindex extends Construct {
 			{
 				lambdaFunction: writeBatchToReindexQueue,
 				inputPath: '$.Payload',
+				payload: TaskInput.fromObject({
+					'input.$': '$',
+					'dryRun.$': '$dryRun',
+				}),
 			},
 		);
 
@@ -195,8 +209,9 @@ export class RecipesReindex extends Construct {
 			);
 
 		// Define the state machine
-		const definition =
-			checkForOtherRunningReindexesTask.next(isOnlyRunningReindex);
+		const definition = storeDryRun
+			.next(checkForOtherRunningReindexesTask)
+			.next(isOnlyRunningReindex);
 
 		// We define the name manually so we can construct the ARN manually and limit the scope
 		// of the ListExecutions permission to the state machine itself - using `stateMachineArn`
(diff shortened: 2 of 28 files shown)


cdk/lib/recipes-backend.ts
```diff
@@ -229,6 +233,21 @@ export class RecipesBackend extends GuStack {
 			],
 		});
 
+		new Rule(this, 'ReindexConnection', {
+			eventBus,
+			description: `Connect recipe responder ${this.stage} to recipes-reindex`,
+			eventPattern: {
+				source: ['recipes-reindex'],
```
jonathonherbert, line 240: Should we be switching on event detail, instead?

cdk/lib/recipes-reindex.ts
```diff
@@ -102,17 +105,38 @@ export class RecipesReindex extends Construct {
...
+					OUTGOING_EVENT_BUS: eventBus.eventBusName,
 				},
 				architecture: Architecture.ARM_64,
 				timeout: Duration.seconds(30),
 				memorySize: 128,
 			},
 		);
 
+		// Necessary to reference `dryRun` later in the step fn.
+		const storeDryRun = new CustomState(scope, 'storeDryRun', {
```
jonathonherbert, line 131: I couldn't find a way of doing this without another node in the step fn graph — but I'd be delighted to find a way.

fredex42:
@jonathonherbert a couple of thoughts from your wonderfully thorough description!

>    the recipe-responder fn currently uses unreserved account concurrency. This means many responder fns can run in parallel. This is not desirable, because it could lead to race conditions as concurrent invocations compete to write their version of the final index to index.json. Perhaps we restrict the responder lambda to a single invocation at a time — or have a separate writer lambda for index.json that has this restriction — to prevent this condition. 

I like the idea of updating the responder lambda so that it does not immediately update `index.json` while re-indexing. It's a relatively expensive operation, requiring a full table scan in DDB and then a dump.  Putting this into a separate phase at the end of step function execution feels neater to me and means we should then be able to parallelise further.

 > Which leads to another thought:  there's no means of handling backpressure outside of AWS's own infrastructure at present. If our batch writer lambda is pushing records faster than downstream can handle them, we don't have visibility of that fact (other than carefully reading logs) — we cannot see whether messages are being dropped, and we cannot clear backlogs as we would be able to in e.g. SQS. Perhaps we can consider adding a queue in front of our responder lambda, which would solve both these problems.

Let's have  a chat about this - i think that adding a queue to manage backpressure is the right approach as it's monitorable and configurable (this is also the approach we take in the search backend, of course)

cdk/lib/recipes-backend.ts
fredex42: I don't think it's necessary in this case - the criterion is that the event is from the recipes reindex process without any other details needed so switching on `source` seems appropriate to me

cdk/lib/recipes-backend.ts
```diff
@@ -229,6 +233,21 @@ export class RecipesBackend extends GuStack {
...
+			},
+			targets: [
+				new aws_events_targets.LambdaFunction(updaterLambda, {
+					deadLetterQueue: responderDLQ,
+					maxEventAge: Duration.minutes(30),
+					retryAttempts: 5,
```
fredex42, line 246: :+1: for including DLQ config.  Though it's worth remembering that this may effectively mix reindex and regular messages on the DLQ. Don't know if that would cause problems yet...

cdk/lib/recipes-reindex.ts
fredex42: :thinking:

lambda/recipes-responder/src/main.ts
```diff
@@ -104,19 +127,57 @@ export async function processRecord({
...
+						contentType: ContentType.ARTICLE,
+						staticBucketName,
+						fastlyApiKey,
+						contentPrefix,
+						outgoingEventBus,
+					});
+				}
+
+				console.log(`Reindexed ${event.detail.articleIds.length} articles`);
+				return totalCount;
```
fredex42, line 173: very minor niggle - to keep with the coding style used in e.g. `processRecord` I'd suggest putting this into its own function?

lib/recipes-data/src/lib/eventbus.ts
```diff
@@ -1,24 +1,23 @@
...
+import type {
+	PutEventsCommandOutput,
+	PutEventsRequestEntry,
+} from '@aws-sdk/client-eventbridge';
 import {
 	EventBridgeClient,
 	PutEventsCommand,
 } from '@aws-sdk/client-eventbridge';
 import { registerMetric } from '@recipes-api/cwmetrics';
-import { OutgoingEventBus } from './config';
```
fredex42, line 7: out of interest - what's the benefit of drilling `OutgoingEventBus` as opposed to pulling it from config here? For testing purposes maybe?

lambda/recipes-responder/src/main.ts
jonathonherbert: e5239ef5f1fc460d94dd9aae54af22984a493f18

lib/recipes-data/src/lib/eventbus.ts
jonathonherbert: Just style, per #86 — now that it's a mandatory property, it avoids runtime surprises. If you'd prefer to avoid drilling we can take another approach — I've found compile-time errors preferable to finding errors in tests, or running lambdas.

fredex42 (approved):
😁 👍

PR #89, opened 2025-01-28
Targeted regions

fredex42:
Co-authored with @Divs-B 

## What does this change?

Implements the first version of regional targeting (https://github.com/guardian/facia-tool/pull/1748) in recipes-backend.
This involves:
- adding the `targetedRegions` and `excludedRegions` fields to the internal data model
- explicitly filtering the Northern hemisphere front based on whether the `us` region is included or excluded
- the Northern hemisphere front will _exclude_ collections specifically targeted at US and _include_ everything else
- a new US-only front is created identical to the Northern hemisphere front that _includes_ collections specifically targeted at US and _excludes_ material specifically targeted elsewhere.
- collections that have no explicit targeting instructions are common to all output fronts

### Future work

Exclusions are not yet implemented in the Fronts tool, so cannot be tested end-to-end in this way.  The exclusion logic is tested in CI though.

## How to test

- Ensure that https://github.com/guardian/facia-tool/pull/1748 is deployed to CODE
- Deploy this to CODE
- Create a front and target some collections to US
- Publish it (in CODE of course!)
- Examine the output data, either in S3 in the CAPI account or via https.
  - https://recipes.code.dev-guardianapis.com/northern/all-recipes/yyyy-mm-dd/curation.json should NOT contain any US targeted containers
  - https://recipes.code.dev-guardianapis.com/us/all-recipes/yyyy-mm-dd/curation.json SHOULD contain any US targeted containers
- Sanity check the new test cases

Able to deliver US targeted content

The biggest risk is that a targeted container displays in the wrong region; this is easily fixable and not a major worry.

### DIFF
--- a/lambda/facia-responder/src/main.test.ts
+++ b/lambda/facia-responder/src/main.test.ts
@@ -47,7 +47,7 @@ describe('main', () => {
 		// @ts-ignore
 		await handler(rec, null, null);
 
-		expect(importCurationDataMock.mock.calls.length).toEqual(2);
+		expect(importCurationDataMock.mock.calls.length).toEqual(4);
 		expect(importCurationDataMock.mock.calls[0][0]).toEqual(
 			JSON.stringify(validMessageContent.fronts['all-recipes']),
 		);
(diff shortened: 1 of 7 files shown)


jonathonherbert (approved):
This all looks good! I haven't had a chance to test — so, approving to unblock, but if you'd like me to exercise the code to double check functionality, let me know and I can get to that on Monday.

lambda/facia-responder/src/main.test.ts
```diff
@@ -70,6 +72,102 @@ describe('main', () => {
...
+					messageId: 'BDB66A64-F095-4F4D-9B6A-135173E262A5',
+					body: JSON.stringify(validMessageUsOnly),
+				},
+			],
+		};
+
+		// @ts-ignore
+		await handler(rec, null, null);
+
+		expect(importCurationDataMock.mock.calls.length).toEqual(4);
```
jonathonherbert, line 121: Nit: I wonder if representing this state as a 2D array or similar might be a bit more readable?

PR #91, opened 2025-02-03
Drop deprecated snyk workflow

fredex42:
## What does this change?

As per advisory from DevX, drop the snyk workflow as it;'s now deprecated and will fail

## How to test

Builds should succeed

See above 😁


Divs-B (approved):
Good to go 👍  
(also as required after https://github.com/guardian/.github/pull/96)

PR #93, opened 2025-03-11
Migrate to Feast account

fredex42:
**NOTE** ~~Do not merge this until the corresponding migration for recipe-search-backend is also ready to go.
The search backend depends on being able to read the messages output from the publication chain, so we must have that available in the same account before go time.~~ DONE

Related recipe-search-backend PR is here: https://github.com/guardian/recipe-search-backend/pull/116

## What does this change?

Migrates the entire recipe-backend over to the Feast account.

This removes the need for a telemetry cross-account role (since telemetry is also migrated) and connects via the secondary event bus defined in https://github.com/guardian/feast-infra

## How to test

- Deploy to CODE using the new `Feast::recipes-backend` name
- Publish a recipe to Feast in CODE while monitoring logs in the Feast space on logs.gutools.co.uk
- It should be picked up and appear in s3 at s3://feast-recipes-static-code, with the index.json getting updated 
- It should also run through the content-api account as well and still be 'live' on the CODE environment
- Repeat the same for launching only to Feast channel
- Take the recipe down again, it should be gone
- Create a new front in Facia tool for the current day
- Publish it
- You should see multiple "published" confirmations - this is because it's publishing to both the 'old' one in content-api and the 'new' one in feast.
- Check s3://feast-recipes-static-code, the front should have come through and be available in the bucket. If it was with todays date it should also be available outside the dated folders

- Able to continue migration
- Unblocking dynamic fronts

We should be able to parallel-run; however, care must be taken during the migration to preserve service.


cdk/lib/recipes-backend.ts
```diff
@@ -83,15 +74,16 @@ export class RecipesBackend extends GuStack {
...
+			description:
+				'ARN of the SNS topic to use for data submissions (shared with structuriser)',
 		});
-		const telemetryTopic = new GuParameter(this, 'TelemetryTopic', {
+
+		const eventBusParam = new GuParameter(this, 'EventBus', {
 			fromSSM: true,
-			default: `/${this.stage}/${this.stack}/${app}/telemetryTopic`,
-			description: 'ARN of the SNS topic to use for data submissions',
+			default: `/${this.stage}/feast/feast-shared-infra/crier-event-bus`,
```
jonathonherbert, line 86: For this and lines above: any reason `this.stack` wouldn't be `feast` in this context?

lib/recipes-data/src/lib/telemetry.ts
```diff
@@ -86,9 +52,8 @@ export async function sendTelemetryEvent(
 		console.error(`sendEvent caught ${typeofErr}: ${JSON.stringify(err)}`);
 		const realAttempt = attempt ?? 1;
 		if (realAttempt < maxAttempts) {
-			console.log('Refreshing credentials and trying again');
+			console.log('Trying again');
```
jonathonherbert, line 55: This log line, seen in isolation, requires you to go to the code to understand it; perhaps we could keep or add to the context we previously had? Or maybe the previous line is incorrect?

jonathonherbert (approved):
IIUC this is out and about in PROD, happy to approve! 🙏  all comments non-blocking

lib/recipes-data/src/lib/telemetry.ts
fredex42: By itself, I agree; however, it would always be preceded by `sendEvent caught ${typeofErr}: ${JSON.stringify(err)}` which is logged at ERROR level so I believe that it does make sense in context:

```
ERROR sendEvent caught someErrorClass: someLongDescription
INFO Trying again
```

jonathonherbert: No guarantee those logs hit ELK in that order, but perhaps this service isn't super chatty, could be fine.

PR #104, opened 2025-04-16
Hybrid container tweaks

fredex42:
## What does this change?

Minor pre-release tweaks to hybrid containers:
- ensure that the localised container does not contain any recipes from the curated front
- remove the `sizeLimit` parameter from the hybrid container fetcher lambda and instead put it at render time (so we can over-request then filter at the end). This is currently hard-coded but simple to change

## How to test

Tested in CI tests

More interesting data!

n/a

### DIFF
--- a/lambda/rest-endpoints/src/curation.ts
+++ b/lambda/rest-endpoints/src/curation.ts
@@ -131,6 +131,7 @@ export async function findRecentLocalisation(
 	cutoffInDays: number,
 	startDate: Date,
 	curatedRecipes?: Set<string>,
+	cutoff?: number,
 ) {
 	for (let i = 0; i < cutoffInDays; i++) {
 		const testDate = subDays(startDate, i);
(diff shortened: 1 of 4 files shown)


lambda/rest-endpoints/src/curation.ts
```diff
@@ -134,12 +141,54 @@ export async function findRecentLocalisation(
...
-			return maybeLocalisation;
+			//if we have been given a set of curated recipes, ensure that they are not in the returned data
+			if (curatedRecipes) {
+				const deduplicatedLocalisation: FeastAppContainer = {
+					...maybeLocalisation,
+					//filter out anything that is a recipe specifier and also exists in curatedRecipes
+					items: maybeLocalisation.items.filter(
+						(row) =>
+							// eslint-disable-next-line no-prototype-builtins -- how else to do this?
+							(row.hasOwnProperty('recipe') &&
```
Divs-B, line 152: Do we need to check if it is `recipe` is it not implicit?

lambda/rest-endpoints/src/curation.ts
```diff
@@ -134,12 +141,54 @@ export async function findRecentLocalisation(
...
+					: maybeLocalisation;
+			}
 		}
 	}
 	return undefined;
 }
 
+function recipeFromContainer(item: ContainerItem): string[] {
+	// eslint-disable-next-line no-prototype-builtins -- how else to do this?
+	if (item.hasOwnProperty('recipe')) {
```
Divs-B, line 180: same here?

Divs-B (commented):
tiny query inline

Divs-B (approved):
Tiny query left otherwise all good 👍

lambda/rest-endpoints/src/curation.ts
fredex42: No, because it can also be `subCollection` or `chef` which support different fields

lambda/rest-endpoints/src/curation.ts
fredex42: No, because it can also be subCollection or chef which support different fields

lambda/rest-endpoints/src/curation.ts
Divs-B: Ah I might have missed it then, I thought it is already a one level deeper where we just have only recipes.
Alright, makes sense 👍

PR #105, opened 2025-04-17
Add json to html renderer for printable version

Divs-B:
## What does this change?

We are providing recipe's print version at server-side, to minimise complexity at client-side apps.

It will need to generate a PDF version of the recipe at publication time, which the app can simply download from the server using a defined url and send to print.

We have used EJS template that will easily render recipe json.
We will be working on docker container with Event Bridge to use this template dynamically for any recipe request.

Trello cards for requirement:
Epic: https://trello.com/c/q5C49kAO/1523-spike-how-to-convert-feast-recipe-page-into-a-printable-pdf?filter=print
Working card: https://trello.com/c/JUGZBHhj/2362-build-server-side-json-html-renderer-for-printable-recipes

## Images

Expected:
image

----------
At present: (Finishing is still left)
image

## Accessibility


Divs-B:
With first commit we have:

image

printable-recipe-generator/src/main.ts
```diff
@@ -0,0 +1,42 @@
...
+
+	//load template
+	const templatePath = path.join(__dirname, 'src', 'assets', 'recipe.ejs');
+	const template = fs.readFileSync(templatePath, 'utf-8');
+
+	//Render html
+	let renderHtml = '';
+	try {
+		renderHtml = renderTemplate(template, recipe as Data);
+	} catch (error) {
```
fredex42, line 21: with my earlier comment about refactoring, I wouldn't `try/catch` here - I'd do it at the outer level and let `renderJsonToHtml` throw as that is more test-friendly

PR #108, opened 2025-05-08
UI-add svgs and fix font for printable recipes

Divs-B:
WIP

## Accessibility


printable-recipe-generator/src/main.ts
```diff
@@ -3,8 +3,63 @@ import * as process from 'node:process';
 import path from 'path';
 import type { Data } from 'ejs';
 import { render as renderTemplate } from 'ejs';
+import fetch from 'node-fetch';
+
+//load Contributors
+async function getChefs(): Promise<unknown> {
```
fredex42, line 9: You shouldn't return a `Promise<unknown>` here really.

The chefs file has a defined format, so you should define each entry as an `interface` and then return a `Record<string, ChefData>` type.

Ideally you'd define it in Zod and use `zod.parse` to validate the data at runtime, but technically it's enough to simply cast the `unknown` value you get back from json (untested code, but it's something like this):

```typescript
  return (await resp.json()) as Record<string, ChefData>
```

printable-recipe-generator/src/main.ts
```diff
@@ -3,8 +3,63 @@ import * as process from 'node:process';
...
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call -- response is expected to be JSON, safe to assign
```
fredex42, line 10: This line should not be needed, assigning `resp` from `fetch` should not cause an error

printable-recipe-generator/src/main.ts
```diff
@@ -3,8 +3,63 @@ import * as process from 'node:process';
...
+	const resp = await fetch(
+		'https://recipes.guardianapis.com/v2/contributors.json',
+	);
+	try {
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-return,@typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-member-access -- returning JSON
```
fredex42, line 15: See my remark above-  you should be able to get rid of this

printable-recipe-generator/src/main.ts
```diff
@@ -3,8 +3,63 @@ import * as process from 'node:process';
 import path from 'path';
 import type { Data } from 'ejs';
 import { render as renderTemplate } from 'ejs';
+import fetch from 'node-fetch';
+
+//load Contributors
+async function getChefs(): Promise<unknown> {
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call -- response is expected to be JSON, safe to assign
+	const resp = await fetch(
+		'https://recipes.guardianapis.com/v2/contributors.json',
```
fredex42, line 12: you should be fetching from CODE environment if this is running in CODE or PROD if it's running in PROD.  Check the environment variable setup in the real lambda - I believe that you should already have a STAGE var set up.

printable-recipe-generator/src/main.ts
```diff
@@ -16,8 +71,15 @@ function renderJsonToHtml(recipeDataPath: string) {
 
 	//Render html
 	let renderHtml = '';
+	// @ts-expect-error -- we know this
```
fredex42, line 74: again, hopefully should not be necessary

printable-recipe-generator/src/main.ts
```diff
@@ -3,8 +3,63 @@ import * as process from 'node:process';
...
 import { render as renderTemplate } from 'ejs';
+import fetch from 'node-fetch';
+
+//load Contributors
+async function getChefs(): Promise<unknown> {
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call -- response is expected to be JSON, safe to assign
+	const resp = await fetch(
+		'https://recipes.guardianapis.com/v2/contributors.json',
+	);
+	try {
```
fredex42, line 14: Personally I wouldn't try/catch here. Instead, I'd allow `resp.json()` to throw an exception and catch it in the caller, something like:

```
let chefs:Record<string, ChefData>|null = null;
try {
  chefs = await loadChefs();
} catch(err) {
  console.error("Could not load chefs: ", err);
}
```

Crucially, you need to remember that `chefs` is now nullable (made explicit by the typing in my example).  Typing it in this way will cause Typescript to remind you of the fact that it _is_ nullable - lines like `console.log(chefs['profile/yotamottolenghi'].webTitle);` will definitely crash if `chefs` is `null` and trying to reference it in the template will likely also cause a carsh

printable-recipe-generator/src/tests/main.test.ts
```diff
@@ -3,11 +3,69 @@ import path from 'path';
 import * as ejs from 'ejs';
 import recipe from '../data/sampleRecipe.json';
 
+//load Contributors
+async function getChefs(): Promise<unknown> {
+	const resp = await fetch(
+		'https://recipes.guardianapis.com/v2/contributors.json',
```
fredex42, line 9: I would not load this in from server in a test - you only want the test to fail if you have a code error, not a failure on the CDN (say).  Instead, I would just create a simple fixture for your chefs object with only the data you need:

```typescript
const chefs:Record<string,ChefData> = {
  'profile/name': {...manually input data...}
}
```

I would also include test cases for three permutations - one where the chef exists in the list, one where the chef does not exist in the list, and one where the list is `null`.  That way we can protect ourselves against unexpected crashes.

printable-recipe-generator/src/tests/main.test.ts
```diff
@@ -3,11 +3,69 @@ import path from 'path';
...
+		.readFileSync(fontPath('GuardianTextSans-Regular.ttf'))
+		.toString('base64'),
+	HeadlineSemiBoldFont: fs
+		.readFileSync(fontPath('GHGuardianHeadline-Semibold.otf'))
+		.toString('base64'),
+	SansBoldFont: fs
+		.readFileSync(fontPath('GuardianTextSans-Bold.ttf'))
+		.toString('base64'),
+};
+
```
fredex42, line 54: I would not duplicate these here. Rather I would put them into their own file - `assetloader.ts`? - and import them here. That way you know that you are getting the same code and you don't need to update it in multiple places.

printable-recipe-generator/src/main.ts
Divs-B: Thanks @fredex42. Yeah thats what I was checking with you that whether it is fine to not to have any interface/type declared and supressing them with ts-ignore tags but it seems we should use interface to avoid those things 👍

printable-recipe-generator/src/main.ts
Divs-B: Fair point, I will change this, but a question: Are we not having CODE and PROD aligned anyway for contributors list?

printable-recipe-generator/src/main.ts
Divs-B: yeah

printable-recipe-generator/src/main.ts
Divs-B: yes changes as mentioned above, resolves this

printable-recipe-generator/src/tests/main.test.ts
Divs-B: O yes, I did this in first place to save writing test data and all and just copied all from main.ts to main.test.ts to pass the test i think. 
Will do the change. thanks.

printable-recipe-generator/src/main.ts
fredex42: No as it's possible that upstream changes could change the data format.... we'd want to be able to test this in CODE (there is a CODE tag manager and CAPI that the data sources from)

printable-recipe-generator/src/main.ts
```diff
@@ -17,7 +48,12 @@ function renderJsonToHtml(recipeDataPath: string) {
 	//Render html
 	let renderHtml = '';
 	try {
-		renderHtml = renderTemplate(template, recipe as Data);
+		renderHtml = renderTemplate(template, {
+			recipe,
+			svgs,
+			fontsBase64,
+			chefs,
+		} as Data);
```
fredex42, line 56: Not sure about the `as Data` cast here - can't see why you'd need to cast?

fredex42 (commented):
Couple more things to test in here, nearly there!

printable-recipe-generator/src/main.ts
```diff
@@ -17,7 +48,12 @@ function renderJsonToHtml(recipeDataPath: string) {
 	//Render html
 	let renderHtml = '';
 	try {
```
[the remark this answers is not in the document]
Divs-B: done

printable-recipe-generator/src/assets/recipe.ejs
```diff
@@ -85,61 +86,98 @@
...
 
-<div class="recipe-title"><%= title %></div>
-<% if (contributors) { %>
-    <div class="author">By <%= contributors %></div>
+<div class="recipe-title"><%= recipe.title %></div>
+<% if (recipe.contributors && typeof chefs === 'object' && chefs?.[recipe.contributors]) { %>
+<div class="author">By <%=  chefs[recipe.contributors].webTitle %></div>
+<% } %>
+<% if (recipe.contributors && typeof chefs != 'object' ) { %>
+    <div class="author">By <%=  recipe.contributors %></div>
```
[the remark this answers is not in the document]
Divs-B: tests added

fredex42 (approved):

PR #110, opened 2025-05-19
To add QR code in recipe pdf

Divs-B:
## What does this change?

We want to facilitate QR codes to the app users on recipe pdf(or printouts). This will let users to open recipe directly on app.
This will enhance user experience too. 
Also, it can make promote new users to download the app via scanning QR codes if they get recipe pdfs and wants to access on app.

QR code will contain the deplink information to open the app
For us it is `feastbraze://recipe/${recipeId}` 
We also putting fallback url for (which I need to confirm with client side)
`https://guardian-feast.go.link/recipe/recipe.id`
ref for trello card: https://trello.com/c/kGg7rZcI/2406-qr-code-in-printable-recipes

Also, PR needs to confirm UI/UX for QR placement in the pdf. At present for testing it is present at the bottom of the screen.

## How to test

Visually QR code should be present
Scan the QR

QR code should open the recipe in the app.


printable-recipe-generator/src/main.ts
```diff
@@ -32,14 +37,20 @@ async function getChefs(): Promise<Record<string, ChefData> | undefined> {
...
 ) {
 	//load recipe JSON
 	const recipe = JSON.parse(
 		fs.readFileSync(recipeDataPath, 'utf-8'),
-	) as unknown;
+	) as RecipeData;
+
+	//load QR Code
+	const recipeId = recipe.id;
+	const deeplinkUrl = `feastbraze://recipe/${recipeId}`;
```
fredex42, line 51: You could just use `feastbraze://recipe/${recipe.id}`

printable-recipe-generator/src/main.ts
```diff
@@ -78,6 +90,6 @@ if (!process.argv[2]) {
 	//Get chefs and render html
 	void (async () => {
 		const chefsList = (await getChefs()) as Record<string, ChefData>;
-		renderJsonToHtml(process.argv[2], chefsList);
+		renderJsonToHtml(process.argv[2], chefsList).catch(console.error);
```
fredex42, line 93: if you're in an async block already you might as well just do `await renderJsonToHtml` rather than using `catch` here. The advantage of _not_ catching is that this should show up as a failed invocation in AWS.

printable-recipe-generator/src/assets/recipe.ejs
```diff
@@ -266,6 +280,17 @@
         <% } %>
     </div>
 </footer>
+
+<%if (qrImageDataUrl && qrImageDataUrl!=null || qrImageDataUrl?.length>0) { %>>
+
+<div class="qr-section">
+    <p>📱 Open in App</p>
+    <img src="<%= qrImageDataUrl %>" alt="Scan to open app" />
+    <p>feastbraze://recipe/<%= recipe.id %></p>
```
fredex42, line 289: Is this necessary? I can't imagine a user ever typing it in. Maybe one for Murray

printable-recipe-generator/src/assets/recipe.ejs
```diff
@@ -266,6 +280,17 @@
...
+    <p>or visit: <a href="https://guardian-feast.go.link/recipe/<%= recipe.id %>">Recipe Here</a></p>
```
fredex42, line 290: this is pointless if you're printing it out! :grin:

Divs-B: Yeah I will be removing these extras 😁

printable-recipe-generator/src/assets/recipe.ejs
Divs-B: Yes, confirmed it with Murray. According to him it will be simple an Image, a QR code with nothing extra details on it,so will remove it.

printable-recipe-generator/src/main.ts
```diff
@@ -32,14 +37,20 @@ async function getChefs(): Promise<Record<string, ChefData> | undefined> {
...
 	const recipe = JSON.parse(
 		fs.readFileSync(recipeDataPath, 'utf-8'),
-	) as unknown;
+	) as RecipeData;
+
+	//load QR Code
+	const recipeId = recipe.id;
+	const deeplinkUrl = `feastbraze://recipe/${recipeId}`;
+	console.log(`deepLInkUrl: ${deeplinkUrl}`);
+	const qrImageDataUrl = await QRCode.toDataURL(deeplinkUrl);
```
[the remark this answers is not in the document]
Divs-B: It will never return null or undefined if it succeeds and If something goes wrong, it will throw an error, not return null.
Do you want me to enclose in try-catch block?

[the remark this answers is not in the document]
fredex42: That should be OK. If it throws, we get a run failure which can be detected in AWS

printable-recipe-generator/src/assets/recipe.ejs
```diff
@@ -46,84 +46,132 @@
...
       }
 
+      /*.two-column-flow {*/
+      /*  display: flex;*/
+      /*  column-count: 2;*/
+      /*  column-gap: 1.5rem;*/
+      /*  !*break-inside: avoid;*!*/
+      /*  !*page-break-inside: avoid;*!*/
+      /*  margin-top: 3rem;*/
+      /*}*/
```
fredex42, line 110: minor - can remove commented code

fredex42 (commented):
one minor tweak and some commented-code zaps then we are good to go 👏

PR #111, opened 2025-06-18
Reindex existing recipe to be pdf ready

Divs-B:
## What does this change?

Writing Node script that should fetch exisitng recipes and send them to Event bridge.
EB should trigger ECS which then start docker container which should then rendering to HTML to generating PDF should happen under the hood.

To run locally, try from root folder:

`tsc -p printable-recipe-generator/tsconfig.app.json` (to compile)
`node dist/out-tsc/printable-recipe-generator/src/triggerReindexPdf.js` (to run script)

This PR has lot of other changes with Lint so running this below shud fix any lint issue as well.
For Prettier run: `npx prettier --write .`

## How to test

All recipes (6259 as of now) should have PDF generated.

ECB invocation should be succesful
LOGS should show PDF
PDF should be open and show valid content.


printable-recipe-generator/docker/docker-entrypoint.sh
```diff
@@ -1,23 +1,14 @@
...
 echo "CHECKSUM ID is $RECIPE_CSID"
-echo "CONTENT is $CONTENT"
-
-
-#Put json content into file
-echo "Write content into a file.."
-RECIPE_CONTENT_PATH="./recipe.json"
-echo "$CONTENT" > "$RECIPE_CONTENT_PATH"
-echo "Recipe json is saved in the path $RECIPE_CONTENT_PATH"
+#echo "CONTENT is $CONTENT"
```
fredex42, line 6:
```suggestion

```

fredex42 (approved):
Just one line to zap

PR #116, opened 2025-07-31
Add metric and alarm for failed dynamic container

Divs-B:
## What does this change?

Recently, we came across the problem where dynamic fronts were not generated since 8th July; I observed on 24th July. 
We didn't have any logs to check on that because it was already over 2 weeks. 
It clearly shows we missed seeing this problem well before.

I raised the concern to Data Tech team, [details in here](https://chat.google.com/room/AAAAAeMn93M/axobtjmPH_8/axobtjmPH_8?cls=10), lambda was not triggered after 8th from Airflow's end and we saw failure Invocation in BQ DAG's too.

Keeping alerts on both ends, one at Airflow's end which can show if the lambda not get triggered and another one at recipe-backend's end where if there is failure then we can have some notification, will help us in dealing with these situations.

James from DT team kindly offered help in checking the divertion of alerts notification to our Feast Dev chat channel.
And This PR change will keep some metric for our end too.

## How to test

- Deploy on CODE
- Observe the metric for `FailedDynamicContainer` after lambda gets triggerd usually by midnight. The metric should be present if there is any failure recorded so we might need to wait if we need to see it.

- Metric should be available and email should be recieved for failed container.

## Accessibility

### DIFF
--- a/cdk/lib/__snapshots__/recipes-backend.test.ts.snap
+++ b/cdk/lib/__snapshots__/recipes-backend.test.ts.snap
@@ -485,6 +485,43 @@ def sort_filter_rules(json_obj):
       },
       "Type": "AWS::CloudWatch::Alarm",
     },
+    "DynamicFrontsFailedDynamicContainerAlarmEAC7CEFB": {
+      "Properties": {
+        "AlarmActions": [
+          {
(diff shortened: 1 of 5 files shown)


alexduf (approved):
Looks good to me! to be merged once we confirm it works on CODE 👍

PR #190, opened 2026-04-15
We are tyring to test manual vs  auto stack update for telemetry topi…

Divs-B:
…c change value

## Accessibility


Divs-B:
It was regarding the deployment error we were seeing on PROD so we wanted to check auto update on AWS stack parameter via riffraff deploy when manual update is done previously. Since the the workaround is done (We replaced updated stack and had to edit the value manually) so we can close this.

PR #197, opened 2026-05-13
Setup devcontainers

fredex42:
## What does this change?

Add setup for devx dev containers

- Check out the branch
- Open in VSCode using 'Open in folder' as normal
- Use Command-Shift-P to open the Command Palette
- Select `Dev containers: Reopen In Container`
- Point to .devcontainer/shared/devcontainer.json
- Let it build and open the container
- Once it's re-opened, open a terminal window
- Run `uname -r` - it'll claim to be Linux!
- Use `npm install`, `npm test`, etc. as per normal
- If the shell complains about not being able to find `npm`, use `mise install` and/or `eval $(mise activate --shims)`

Road-testing a more secure development process

### DIFF
--- a/.devcontainer/.gitignore
+++ b/.devcontainer/.gitignore
@@ -1,0 +1,2 @@
+# User-specific devcontainer directory with merged personal preferences
+user/
--- a/.devcontainer/devenv.yaml
+++ b/.devcontainer/devenv.yaml
@@ -1,0 +1,40 @@
+# Devenv project configuration
+# Edit this file to configure your project's devcontainer
+# and then run `devenv generate` to create the devcontainer.json files
+
+# REQUIRED: Change this to your project name
+name: "recipes-backend"
+
+# Modules: Built-in functionality
+# - mise: Install and configure mise for dev tools management (https://mise.jdx.dev/)
+# - docker-in-docker: Enable running Docker containers within the devcontainer
+# - scala: Add IDE plugins and jar caching for Scala development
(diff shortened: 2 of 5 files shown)


Divs-B:
I ran through the instructions and installed Mise and Node for the setup of Devcontainers in VS Code.

image

Divs-B:
A question: I can see VS code naturally fitting for Dev Containers, I am curious to check IntelliJ setup for same? 
Means talking to Docker, build the container, and open up a workspace running inside that secure environment—giving us the exact same mise and Node setup? What's your opinion for future development for IDE wise?

Divs-B (approved):
The instructions were followed, and got things running Visual Studio Code.
I have one question left, and I would be happy to approve it to unblock it.

fredex42:
@Divs-B it also works fine in IntelliJ - I'm doing Unified API work and recipe search work under dev containers in IntelliJ too.  The process is a little different - you need to get to the "welcome" screen, then select Dev Containers, and there is an option to create a dev container from either a local repo or a remote one.  After that it all works the same - builds the container, installs the stuff, runs it and provides terminal access etc. in the UI

It's not _quite_ as reliable - few rough edges still (I've had to manually initiate installs and fix PATH issues before) but it does work.

What's nice is that it gives you all the right versions of languages and frameworks out of the box - as well as improving isolation from supply chain attacks

@adamnfish-gu @aug24 @tjsilver can provide with more information on running in intellij/vscode 😁 

Many thanks for the ✅

PR #203, opened 2026-06-11
Cmd tool for terminology csv-json-s3

Divs-B:
## What does this change?

This is command line tool to use a Typescript based script that reads CSV file, converts it into JSON and put it in S3 bucket.

In the initial phase of terminology development, we will utilise a CSV file for one-to-one mapping across the entire recipe for US-based users. This approach will provide a straightforward and efficient method for this task. Should the requirement for this task increase, we may consider implementing a web tool that will allow editorial team members to perform this task independently.

CSV file at present is in the format of id, ukTerm, usTerm

npm install
Put feast key
export AWS_PROFILE=feast

Then run these commands from inside the update-terminologies folder
To see list of files -->>
npx tsx csv-json-s3.ts list --stage CODE

To update the file on S3 bucket -->> 
npx tsx csv-json-s3.ts update --stage CODE --file ‘<file>’

To rollback -->>
npx tsx csv-json-s3.ts rollback --stage CODE --datetime <date>

These commnds should work as expected.

## Accessibility

### DIFF
--- a/tools/update-terminologies/csv-json-s3.ts
+++ b/tools/update-terminologies/csv-json-s3.ts
@@ -1,0 +1,265 @@
+#!/usr/bin/env node
+/**
+ * terminologies tool — convert a UK/US terminologies CSV to JSON and manage it in S3.
+ *
+ * CSV input has 3 columns: id, ukTerm, usTerm
+ *
+ * Produces JSON of the shape:
+ *   {
+ *     "prepared_at": "2026-06-02T10:30:00Z",
+ *     "key": ["id", "ukTerm", "usTerm"],
+ *     "values": [[1, "sugar", "sugar"], ...]
+ *   }
+ *
+ * S3 layout (per stage bucket):
+ *   terminologies/terminologies-<timestamp>.json   <- dated archives (what `list` shows)
+ *   terminologies/latest/terminologies.json        <- live file the library reads
+ *
+ * Commands:
+ *   update   --stage CODE --file ./terms.csv
+ *            -> writes a new dated archive AND overwrites the live file
+ *   list     --stage CODE
+ *            -> lists the dated archives in the terminologies folder
+ *   rollback --stage CODE --datetime 2026-06-02T10-30-00
+ *            -> copies the matching archive back over the live file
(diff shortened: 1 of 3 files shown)


tools/update-terminologies/csv-json-s3.ts
```diff
@@ -0,0 +1,248 @@
...
+
+function archiveKey(stamp: string): string {
+	return `${ARCHIVE_PREFIX}${stamp}/terminologies.json`;
+}
+
+// id should be a number; fall back to the raw string if it isn't numeric.
+function toId(raw: string): number | string {
+	const n = Number(raw);
+	return Number.isNaN(n) ? raw : n;
+}
```
fredex42, line 57: Technically I doubt if the ids are necessary but there's no harm in having them

fredex42 (commented):
Looks good to me.

One point - you're duplicating the 'stage' input, as both an environment variable and a CLI argument.

I'd drop the env var in favour of the CLI argument; you should also drop the BUCKET env var, because if the BUCKET is set to the PROD bucket but you run with `--code` you'd expect the app to hit CODE.

I'd simply use `const bucket = 'feast-recipes-static-{params.stage.toLowerCase()}'` (with the right quote marks I can't type here!) 😁

tools/update-terminologies/csv-json-s3.ts
Divs-B: Yeah I thought lets have decent structure so started with `id` lol

Divs-B:
> Looks good to me.
> 
> One point - you're duplicating the 'stage' input, as both an environment variable and a CLI argument.
> 
> I'd drop the env var in favour of the CLI argument; you should also drop the BUCKET env var, because if the BUCKET is set to the PROD bucket but you run with `--code` you'd expect the app to hit CODE.
> 
> I'd simply use `const bucket = 'feast-recipes-static-{params.stage.toLowerCase()}'` (with the right quote marks I can't type here!) 😁

Thanks for checking, I missed out removing env variable one.

fredex42 (approved):

PR #204, opened 2026-06-11
App exclusive branding field

fredex42:
## What does this change?

- Adds "appExclusiveBranding" field to the RecipeV3 definition as an optional field. This will be provided by the corresponding Composer PR at https://github.com/guardian/flexible-content/pull/6609.

- Deploy the above Composer branch to CODE
- Deploy this to CODE
- Open a recipe element in Composer CODE (as per the instructions on the above PR)
- Set the "app exclusive branding" flag, launch the recipe to CODE
- Find the recipe ID via the "View JSON" functionality on the recipe element in Composer
- Resolve the recipe in the recipe backend: https://recipes.code.dev-guardianapis.com/api/content/by-uid/{recipe-id-here}
- You'll see the new field in the output

- Able to unblock app exclusive recipe branding
- Able to update the kotlin runtime library

### DIFF
--- a/schema/recipe-v3.json
+++ b/schema/recipe-v3.json
@@ -8,6 +8,10 @@
     "id"
   ],
   "properties": {
+    "appExclusiveBranding": {
+      "type": ["boolean", "null"],
+      "description": "If set, render the recipe with app-exclusive branding and styling. If not present, must be treated as false."
+    },
     "isAppReady": {
       "type": "boolean",
       "description": "Whether the recipe is ready for app display"
--- a/schema/recipe-v3.json
+++ b/schema/recipe-v3.json
@@ -9,8 +9,8 @@
   ],
   "properties": {
     "appExclusiveBranding": {
-      "type": ["boolean", "null"],
-      "description": "If set, render the recipe with app-exclusive branding and styling. If not present, must be treated as false."
+      "type": "boolean",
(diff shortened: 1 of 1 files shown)


schema/recipe-v3.json
```diff
@@ -8,6 +8,10 @@
     "id"
   ],
   "properties": {
+    "appExclusiveBranding": {
+      "type": "boolean",
+      "description": "If true, render the recipe with app-exclusive branding and styling. If not present, must be treated as false."
```
Divs-B, line 13: Where are the decisions regarding branding and styling components residing? Are they at the application level, or is this something that the server will need to consider in the future? Do you know any thing on this matter at this time?

schema/recipe-v3.json
```diff
@@ -8,6 +8,10 @@
     "id"
   ],
   "properties": {
+    "appExclusiveBranding": {
```
Divs-B, line 11: this must be optional from downstream to not break anything upstream ya?

Divs-B (approved):
A few queries for the sake of logistics, but the process is straightforward and easy to approve so 👍

schema/recipe-v3.json
fredex42: I don't.  That will have to come in time 🤷

schema/recipe-v3.json
fredex42: yep. It's simple enough, fortunately, to treat optional === false

PR #214, opened 2026-06-23
add "original measuring system" to the recipe element definition

fredex42:
## What does this change?

Adds the "original measuring system" field from https://github.com/guardian/flexible-content/pull/6633

- Deploy the Composer PR
- Publish a recipe with "original measuring system" set
- Validate it strictly against the JSON schema

Able to use the field

### DIFF
--- a/schema/recipe-v3.json
+++ b/schema/recipe-v3.json
@@ -115,6 +115,10 @@
         "$ref": "ingredients-list.json"
       }
     },
+    "originalMeasuringSystem": {
+      "enum": ["metric", "us", "aus-cup", "imperial"],
+      "description": "The original measuring system used in the recipe. This is a view hint for conversion and should not be taken as a strict source of truth."
+    },
     "celebrationIds": {
       "type": [
         "array",


schema/recipe-v3.json
```diff
@@ -115,6 +115,10 @@
         "$ref": "ingredients-list.json"
       }
     },
+    "originalMeasuringSystem": {
```
Divs-B, line 118: It is certainly worth considering this inclusion. I had also suggested a similar approach when working in terminology. This has the potential to address a wide range of conversion scenarios, including terminology-related challenges in future.

Divs-B, line 118: Suggestion: The enum appears to be a combination of regional and measuring types. To provide more scenarios, we could separate the regions and measuring systems (mathematical) separately. However, we can leave it for future development.  

```{
  "type": "object",
  "properties": {
    "measuringSystem": {
      "type": "string",
      "enum": ["metric", "us_customary", "imperial"],
      "description": "The mathematical system of units used for the recipe's base measurements (e.g., metric handles grams/ml, us_customary handles cups/ounces)."
    },
    "targetRegion": {
      "type": "string",
      "enum": ["global", "us", "uk", "au", "ca"],
      "description": "The specific geographic region the recipe was written for. This resolves localized variations within a system, such as an Australian cup (250ml) vs a US cup (240ml)."
    }
  },
  "required": ["measuringSystem", "targetRegion"]
}

Divs-B (approved):
👍

schema/recipe-v3.json
fredex42: Yes I considered this - it's a bit complex from the user's point of view though. Something to revisit in the future though.

fredex42: e.g. we could use just a targetRegion, and treat each as having its own "customary" unit system - that would be the most flexible - but hellishly fiddly to get right!

PR #215, opened 2026-06-26
Package updates

fredex42:
## What does this change?

Updates vulnerable packages.

> [!NOTE]
> We can't update Thrift at the moment as it's pinned by upstream library dependencies.  The data we parse isn't really "external" but this will need closer co-ordination with the upstream content pipeline

- Deploy to CODE
- Publish recipes
- Check they come through


Divs-B (approved):
Based on CI/CD passing and your tests, 👍

PR #220, opened 2026-07-29
slightly loosen image url regex so valid images don't get rejected

fredex42:
## What does this change?

Fixes an issue reported where one image was not getting the proper Fastly scaling URL when published.

The issue turned out to be that the URL used to validate and extract the data from the Grid URL was overly strict, only matching between 1 and 4 digits in the encoded data part.  The affected image had 5 digits in the last (unused) part.

The PR updates the regex to allow up to 6 digits in the unknown parts and adds a test for the specific problematic URL to avoid a regression.

- Deployed to CODE
- Relaunched https://composer.code.dev-gutools.co.uk/content/6a2fdf5f8f084fe6123b53f4
- Check that the image URL in the resulting recipe on [recipes.guardianapis.com](https://recipes.code.dev-guardianapis.com/api/content/by-uid/01ddc030f3c94c11ad0a7df8d3055a14) is of the form `https://i.guim.co.uk/img/media/....` not of the form `https://media.guim.co.uk/....` - once the CDN caching clears!

All images displaying properly in the app

n/a

### DIFF
--- a/lib/recipes-data/src/lib/utils.test.ts
+++ b/lib/recipes-data/src/lib/utils.test.ts
@@ -131,6 +131,14 @@ describe('extractCropIdFromGuimUrl', () => {
 			4259,
 			'png',
 		],
+		[
+			//This specific URL form was causing problems since the last number was too long, so we've added a test case to ensure it doesn't regress
+			'https://media.guim.co.uk/6a4de89b071368602094005335a3266910a36f0a/675_755_9267_11585/1600.jpg',
+			'6a4de89b071368602094005335a3266910a36f0a',
+			'675_755_9267_11585',
+			9267,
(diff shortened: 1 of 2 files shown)


Divs-B:
To test the CODE, I have obtained a version of the composer link for the problem recipe (with a pixelated image).
https://composer.code.dev-gutools.co.uk/content/6a2fdf5f8f084fe6123b53f4
I requested Laura from the Editorial team to confirm the improved clarity of the image, but she was unable to do so as she does not have access to CODE Composer.

As @fredex42 mentioned to us, the code is currently working, so I am happy to unblock it.

Divs-B (approved):
Thanks for fixing 👍

